PR #23, opened 2024-05-29
fix: auth_client token generation

mt-czi:
The `token_prepare_request` [method](https://github.com/oauthlib/oauthlib/blob/master/oauthlib/oauth2/rfc6749/clients/base.py#L260) for oauth library fails. The code is passing the authorization url without the attributes containing the code value. This raises an error from the oauth library complaining no code was passed. However we do do pass the actual code as a separate parameter. So removing the authorization_url parameter fixes the issue.


kharkevich:
The issue was closed as stale due to prolonged inactivity

PR #46, opened 2024-10-24
feat: check access token for group attribute as well; fix #43

### LINKED ISSUE #43: Check Access Token for OIDC_GROUPS_ATTRIBUTE
Hi!

Thanks a lot for this plugin! I was trying it with my OAuth IP and in my specific case the attribute (`OIDC_GROUPS_ATTRIBUTE`) is part of the access token and not of the user token. I made some modifications in the code to check the access token as well. Are you interested in PR? I could also add a plugin but I was thinking that this use case might be common enough to include in the main code.

hahahannes:
This will add the feature of checking the access token for the `OIDC_GROUPS_ATTRIBUTE`. This is useful for cases where the needed value is stored in the access token and not in the user info token

### DIFF
--- a/mlflow_oidc_auth/config.py
+++ b/mlflow_oidc_auth/config.py
@@ -34,6 +34,7 @@ class AppConfig:
     OIDC_REDIRECT_URI = os.environ.get("OIDC_REDIRECT_URI", None)
     OIDC_CLIENT_ID = os.environ.get("OIDC_CLIENT_ID", None)
     OIDC_CLIENT_SECRET = os.environ.get("OIDC_CLIENT_SECRET", None)
+    OIDC_AUDIENCE = os.environ.get("OIDC_AUDIENCE", None)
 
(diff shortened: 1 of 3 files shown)


kharkevich:
@hahahannes thanks for your PR, could you please check #54 probably it will work for you. Otherwise, please update your PR in accordance with the new project structure

NeroBlackstone:
This problem still exists in the latest 3.0 version, maybe you need to re-fork and fix it on the latest branch?
(This is a useful fix, great job

hahahannes:
Hi @NeroBlackstone @kharkevich,
in the meantime, in my project we got the roles in the `userinfo` object. So for me, there is no need to get it from the actual `access_token` anymore. 

@NeroBlackstone You can write a `OIDC_GROUP_DETECTION_PLUGIN` as this gets the actual accesss_token as a parameter and then you can check for your custom field

PR #47, opened 2024-10-24
feat: token authentication for api access; fix #44

### LINKED ISSUE #43: Check Access Token for OIDC_GROUPS_ATTRIBUTE

### LINKED ISSUE #44: Validate access token for authentication
At the moment only basic authentication is performed. Additionally I would like to validate the access token for authentication if one is present. I have code for this. Let me know if you are interested in a PR.

hahahannes:
This Pr adds the feature of validating an OAuth/OIDC access token in the `Authorization` header for authentication. First the presence of a token is checked, then its validity. In case of failure, the original basic auth check is performed. At the moment, I tested it with Keycloak as the Identity Provider and RS256 signed tokens and the public keys that are exposed by keyloack under `/protocol/openid-connect/certs`

### DIFF
--- a/mlflow_oidc_auth/config.py
+++ b/mlflow_oidc_auth/config.py
@@ -34,6 +34,11 @@ class AppConfig:
     OIDC_REDIRECT_URI = os.environ.get("OIDC_REDIRECT_URI", None)
     OIDC_CLIENT_ID = os.environ.get("OIDC_CLIENT_ID", None)
     OIDC_CLIENT_SECRET = os.environ.get("OIDC_CLIENT_SECRET", None)
+    OIDC_AUDIENCE = os.environ.get("OIDC_AUDIENCE", None)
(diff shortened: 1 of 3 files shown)


hahahannes:
@kharkevich Let me know what you think. I tried to make it as generic as possible but it needs more testing for the full feature. Could definitely also add some unittest.

malleshasc (approved):
Approving request, this code would help us.

kharkevich:
@hahahannes thanks for inspiring me to make this feature real, could you please check #54 probably it will work for you. Otherwise, please update your PR in accordance with the new project structure

hahahannes:
Awesome! I will try to test it soon

hahahannes:
Closing this, as the recent update seems to work for me

PR #54, opened 2024-12-07
feat: refactoring

kharkevich:
BREAKING CHANGE: OIDC configuration changed. Only discovery URL supported now


mshodge:
Hi @kharkevich just checking, does this mean OIDC_AUTHORIZATION_URL, OIDC_TOKEN_URL and OIDC_USER_URL are no longer supported? I was building something using GitHub last week but that doesn't use DISCOVERY_URL method

mlflow_oidc_auth/auth.py
```diff
@@ -0,0 +1,63 @@
...
+    metadata = requests.get(config.OIDC_DISCOVERY_URL).json()
+    jwks_uri = metadata.get("jwks_uri")
+    jwks = requests.get(jwks_uri).json()
+    cache.set("jwks", jwks, timeout=3600)
+    return jwks
+
+
+def validate_token(token):
+    jwks = _get_oidc_jwks()
+    payload = jwt.decode(token, jwks)
```
hahahannes, line 38: I think I had problems here that the library needed the `audience` parameter if the token has an `aud` field. Just mentioning, haven't tested the latest version yet

PR #56, opened 2024-12-11
feat: add `OIDC_ALLOW_ALL_USERS` , allow all users to access, even those not in the allowed groups.

NeroBlackstone:
In this PR:

- Add OIDC_ALLOW_ALL_USERS to allow all users to access. 
Motivation: Assuming there is a self-hosted gitlab in the organization, all users in gitlab should have authorized to access MLFlow. Save the trouble of manually adding them to the group one by one

- Added a script to initialize the development environment.(clear authorized user data)
- Add `strtobool` function, avoiding dependencies on distutils
- Avoid warnings: `Usage of '~' in imports is deprecated.`
- Fix bugs: If there is no `groups` field in the `id token` and `userinfo`, HTTP 500 will be caused


NeroBlackstone:
If you find it inconvenient to review, I can close this PR and split it into other PRs (I think the code changes will be clearer this way).

kharkevich:
dear @NeroBlackstone,
could you please split this into separate PRs?

regarding the GitLab integration case, could you check if it can be implemented as an independent plugin (similar to how group retrieval works for MS Entra ID)?

NeroBlackstone:
Close PR for split it into other PRs

PR #61, opened 2024-12-16
fix: ui components missing in release

ion-elgreco:
Some UI components were missing in the wheel. This is because the angular max error size was too low so it failed building, however the CI script was also not strict enough, so it was able to continue the other commands even though exit code 1 was return on the yarn build command.

Since https://github.com/data-platform-hq/mlflow-oidc-auth/pull/54 menu.html also wasn't included anymore during the build

- closes https://github.com/data-platform-hq/mlflow-oidc-auth/issues/60
- closes https://github.com/data-platform-hq/mlflow-oidc-auth/issues/49


ion-elgreco:
@kharkevich could you push a fix release with this? V2.0+ release are currently broken :(

kharkevich (approved):

PR #66, opened 2025-01-08
check access token for group membership; fix https://github.com/data-…

hahahannes:
…platform-hq/mlflow-oidc-auth/issues/43


hahahannes:
What do you think about this @kharkevich ? Similar to the other user, we have the case where the groups value is placed in the access token itself and not in  the user id token.

hahahannes:
I realized that I should use the `authlib` and not the `jwt` package. Will change

hahahannes:
@kharkevich Ready for review :)

hahahannes:
Hi, closing this for now. In the end using the user token worked for me. So there was no need for this feature anymore. If needed, I can work on a new one based on the latest release.

PR #70, opened 2025-01-08
feat: use trusted header to get the user id instead of authentication

### LINKED ISSUE #69: Use user id from a trusted header in case mlflow runs behind proxy
Current State:
Authentication is done completely by the plugin.

New feature:
Having the possibility to use the user id from a header and to bypass the plugin authentication

My use case:
I am running MlFlow behind a proxy which already handles authentication. I can pass a header with the user email downstream to the plugin. It would be nice to use it. I know this plugin is mainly about OIDC authentication but I also really like the authorization feature.

hahahannes:
This will add the possibility to trust a custom header providing the user id. In one of our use cases, this is useful when MlFlow runs behind a reverse proxy which is responsible for the authentication.

This PR is in progress. For example it is not clear how the first request of a user should be handled as the user has to be first created in the database.

Fixes https://github.com/mlflow-oidc/mlflow-oidc-auth/issues/69

### DIFF
--- a/mlflow_oidc_auth/auth.py
+++ b/mlflow_oidc_auth/auth.py
@@ -8,6 +8,8 @@ from werkzeug.datastructures import Authorization
 
 from mlflow_oidc_auth.config import config
 from mlflow_oidc_auth.store import store
+from mlflow_oidc_auth.user import create_user 
+from mlflow.exceptions import MlflowException
 
 
 _oauth_instance: Optional[OAuth] = None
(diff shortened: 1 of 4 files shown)


hahahannes:
I think this is still an interesting feature for running MlFlow behind reverse proxies that handle authentication (In my case, e.g. the proxy also allows kubernetes authentication which is a cool feature if your connecting to mlflow from within a Kubernetes cluster). I assume this is not in scope for this plugin as it focuses on oidc authentication but I really the UI and the permission handling, so I thought, it could make this plugin more towards a general purpose auth plugin.

kharkevich:
Thanks @hahahannes — reverse-proxy/K8s header auth is a legitimate use case we'd like to support. Closing this particular branch for two reasons: (1) it targets the pre-FastAPI `legacy` architecture and would need a full rewrite as middleware; (2) as written the header is trusted unconditionally, with no verification that the request actually came from the trusted proxy — a client reaching the server directly could send the header and be authenticated as anyone. A safe implementation needs proxy-source enforcement (building on the existing `TRUSTED_PROXIES` config), inbound-header stripping, and controlled provisioning. We'll track this as a fresh feature; your PR is the reference for the requirement. Not a reflection on the effort — the security model just needs to be built in from the start.

Issue #76, opened 2025-02-27
[FEAT] Using the same database for both mlflow and mlflow-oidc-auth not possible due to sharing alembic version tables


gdiepen:
In my current situation, I do not have the option of creating two separate databases for both mlflow and mlflow-oidc-auth. 

I just have only one postgres database available. When I point both mlflow and mlflow-oidc-auth to this same database, the upgrades of both applications will go wrong because they both encounter a version they do not know from the other application.

Would be good if it is possible to use an environment variable to use a non-default name for the version_table.

Will create a PR that implements exactly this

hahahannes:
Hi! I am using two separate databases within one postgres instance. Just as a quick comment :)

gdiepen:
I understand, but in the situation at my work I can only generate one managed database for my application and do not have an option of creating two separate databases myself within one postgres instance

PR #77, opened 2025-02-27
feat: Make alembic version table configurable

gdiepen:
If the environment variable `OIDC_ALEMBIC_VERSION_TABLE` is defined, use this as the table that alembic uses to store the alembic versions in. If not defined, use the default `alembic_version`.

Fixes: #76

Allows to use both mlflow-oidc-auth and mlflow in the same (postgres) database, each having its own alembic version table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ The plugin required the following environment variables but also supported `.env
 | SECRET_KEY             | Key to perform cookie encryption |
 | LOG_LEVEL                   | Application log level |
 | OIDC_USERS_DB_URI | Database connection string |
+| OIDC_ALEMBIC_VERSION_TABLE  | Name of the table to use for alembic versions (defaults to alembic_version if not provided)                                                          |
 
 ## Application session storage configuration
 | Parameter | Description | Default |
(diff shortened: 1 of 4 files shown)


gdiepen:
Will check what needs to be changed to use mlflow_oidc_auth/config.py there

Also will try to see how to write the test for this

gdiepen:
So the first part is done, I have moved everything to the config and use that.

Second part I am having some problems with, not familiar with python unittest  (a bit more experience with pytest). 

Getting stuck on how to ensure I can initialize an empty database and see if the table created is based on the os environment.....

gdiepen:
@kharkevich If you understand what is going wrong here, that would be very useful

If I use `:memory:` for the create_engine function in line 49, like this
image

the tables list is always empty.

However, if I change it to `engine = create_engine("sqlite:///foo.db")`, then it does work.

Second question would be how to mock environment variable in this framework, such that I can create a second test where I let it be a custom alembic version table

gdiepen:
@kharkevich Included unit tests for the new functionality. Had to rely on writing actual databases to some temporary files to make the checks actually work.... But the tests are now working

gdiepen:
The unit tests for this one kinda works, but not completely. 

The problem is that the AppConfig object is only created once, which means it is difficult to mock that based on different environment variables because changes are not picked up....

Do you have any idea about how you want to deal with this?

gdiepen:
I could not do the patching/mocking of the app config object in the env module, because that would already load it in a different way and the context would not be available. Therefore done in the current way during the unit tests

Issue #78, opened 2025-02-27
Use patterns for group names allowed to login to mlflow


gdiepen:
In the current setup the dealing with the groups might not always be straight forward

In the case you are using the standard okto way for retrieiving all groups, it could be that if the user is a member of 200 groups, we add 200 groups to the authorization database which might not be useful.

Also, if we use the plugin to create a multi-tenant setup and we would add a new team that also would have access to the mlflow server, this would require a modification in the application configuration because we would need to add this new group to the environment / .env file

Instead of having explicit group names, we could also use glob patterns or regular expressions that we use to keep only those group names that match the patterns. If after the filtering, the list of groups is empty it means the user is not part of any group-pattern and thus should be denied access.

This way, if the group names follow a certain pattern, we can leverage that and do not have to update any configuration when new teams are introduced. E.g. if we have groups like `mlflow-team1-datascientist`, mlflow-team1-modelpublishers`, mlflow-team2-datascientist`, etc we could create a new application configuration to be a space/comma separate list of either glob patterns or regex strings

```
# Glob pattern for each team separate, separated by comma
OIDC_GROUP_FILTERS=mlflow-team1-*,mlflow-team2-*

# glob pattern that matches all teams at once, assuming that everything starting with mlflow- should give access
OIDC_GROUP_FILTERS=mlflow-*
```

The highlevel code for the user_groups related part in `views/authentication.py` could then become something like:
```python
import fnmatch

# Could come from config.OIDC_GROUP_FILTERS
FILTERS = "mlflow-*,datascience-*"

# Suppose this comes from the token["userinfo"][config.OIDC_GROUPS_ATTRIBUTE]
user_groups = [
    "mlflow-team1-datascience",
    "mlflow-admin",
    "other-application-role",
    "other-application2-role",
    "datascience-member",
]

# Filter the user groups
user_groups = sorted(
    set(
        [
            x
            for f in FILTERS.split(",")
            for x in [g for g in groups if fnmatch.fnmatch(g, f)]
        ]
    )
)

# Prints: ['datascience-member', 'mlflow-admin', 'mlflow-team1-datascience'] for the example data
app.logger.debug(f"User groups: {user_groups}")

if not len(user_groups):
    return "User is not allowed to login", 401
``` 

This way any new groups that are added following the GROUP_FILTERS pattern automatically will be allowed to login to the system and we only create groups in the auth database for the relevant groups this person is a member of.

[...]

PR #81, opened 2025-03-04
feat: use patterns to filter all groups to only relevant groups

gdiepen:
In the original setup you have a configuration option `OIDC_GROUP_NAME` which provides a comma separated list of groups that are considered relevant. 

Every new group that is added to AD that would allow access to MLFlow would require a restart of the mlflow server with an updated configuration.

Instead, if we use patterns like mlflow-users-* as an allowed pattern, it would indicate that any group starting withmlflow-users- would be a group that gives access to mlflow.

One additional benefit is that if the okta function returns a large number of groups (e.g. 300 different groups that a user is a member of), only the groups considered relevant based on the patterns are added to the permission databases, the rest is ignored.

This PR modifies the code to not work with a hardcoded list of full names, but with a list of patterns instead. 

Decided to not use full regular expressions here, but simple file-system like matching using the `fnmatch` library in python.

Implements: #78

### DIFF
--- a/mlflow_oidc_auth/config.py
+++ b/mlflow_oidc_auth/config.py
@@ -19,7 +19,9 @@ class AppConfig:
         self.DEFAULT_MLFLOW_PERMISSION = os.environ.get("DEFAULT_MLFLOW_PERMISSION", "MANAGE")
         self.SECRET_KEY = os.environ.get("SECRET_KEY", secrets.token_hex(16))
         self.OIDC_USERS_DB_URI = os.environ.get("OIDC_USERS_DB_URI", "sqlite:///auth.db")
-        self.OIDC_GROUP_FILTER_PATTERNS = [pattern.strip() for pattern in os.environ.get("OIDC_GROUP_FILTER_PATTERNS", "*").split(",")]
+        self.OIDC_GROUP_FILTER_PATTERNS = [
+            pattern.strip() for pattern in os.environ.get("OIDC_GROUP_FILTER_PATTERNS", "*").split(",")
+        ]
         self.OIDC_ADMIN_GROUP_NAME = os.environ.get("OIDC_ADMIN_GROUP_NAME", "mlflow-admin")
         self.OIDC_PROVIDER_DISPLAY_NAME = os.environ.get("OIDC_PROVIDER_DISPLAY_NAME", "Login with OIDC")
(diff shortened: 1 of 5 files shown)


gdiepen:
I am currently working on the unit tests. Having some problems with the fact that the AppConfig object is evaluated only once, trying to work around this.

gdiepen:
@kharkevich This feature is fully implemented now, including the test functions

grudloffev:
Hi, I believe this PR is really useful, however I don't think it's a good idea to break backwards compatibility by changing the names

gdiepen:
I explicitly did not want to re-use the existing name because the logic is very different (in essence it is a filter instead of just a list).

Let me know what you think is the best solution and happy to modify the code.

grudloffev:
hmm so it only allows filters? Wouldn't it make sense to support both in the same list and handle it accordingly?

FYI I am not a maintainer, so just voicing my personal opinion here.

mlflow_oidc_auth/config.py
```diff
@@ -19,7 +19,9 @@ def __init__(self):
         self.DEFAULT_MLFLOW_PERMISSION = os.environ.get("DEFAULT_MLFLOW_PERMISSION", "MANAGE")
         self.SECRET_KEY = os.environ.get("SECRET_KEY", secrets.token_hex(16))
         self.OIDC_USERS_DB_URI = os.environ.get("OIDC_USERS_DB_URI", "sqlite:///auth.db")
-        self.OIDC_GROUP_NAME = [group.strip() for group in os.environ.get("OIDC_GROUP_NAME", "mlflow").split(",")]
+        self.OIDC_GROUP_FILTER_PATTERNS = [
+            pattern.strip() for pattern in os.environ.get("OIDC_GROUP_FILTER_PATTERNS", "*").split(",")
```
jackylamhk, line 23: Using `str.split(",")` here may break expressions that include commas. Would parsing a JSON array be more robust?

kharkevich:
Thanks @gdiepen — pattern-based group filtering (#78) is a useful capability and still missing on `main`. Closing because the `legacy` base branch is being decommissioned and this diff targets the old layout. Issue #78 remains open for a `main`-targeted reimplementation. Not a won't-fix — a wrong-target retirement.

Issue #83, opened 2025-03-07
[FEAT] Default per group permission


grudloffev:
I believe it would be really useful to have the option to set a default permission over groups, much like the `DEFAULT_MLFLOW_PERMISSION` environment variable right now. This parameter could be set through an environment variable too, for instance `DEFAULT_MLFLOW_GROUP_PERMISSION` and would have the effect of setting the default permissions to that value over the members of the group.

This would, for example, permit the usecase of granting joint access to members within a group while isolating access between different groups. This would be achieved with the following setting:
```
DEFAULT_MLFLOW_PERMISSION="NO_PERMISSIONS"
DEFAULT_MLFLOW_GROUP_PERMISSION="READ"
```

grudloffev:
Below is an illustration of what this feature allows with:
-  `DEFAULT_MLFLOW_PERMISSION="NO_PERMISSIONS"`
-  `DEFAULT_MLFLOW_GROUP_PERMISSION="READ"`
- `my_user` in one group, while `my_userother1` and `my_userother2` in another group.
- `my_usernoauth` does not belong to any group.
- `my_admin` belongs to the admin group
- Each user creates an experiment

<img width="633" height="502" alt="Image" src="https://github.com/user-attachments/assets/33f152d1-5559-475a-8cec-852b5dee3507" />

<img width="633" height="502" alt="Image" src="https://github.com/user-attachments/assets/7e173a1e-34e6-4947-a15d-622a38d281e4" />

kharkevich:
Unfortunately, it might be work for some simple cases when your user have no complicated group membership interspersions,
for example: 

- user1 is member of: group1, group2, group3, group5
- user2 is member of: group2, group4, group5
- user3, is member of: group1
- user4 is member of: group2
- user5 is member of: group3
- user6 is member of: group4
- user7 is member of: group5

each group is used to separate access to sensitive experiment data,

user1 create experiment and automatically break security requirements, 

to solve this issues I suggest to use separation based on naming conventions and regular expressions to grant default access to user directly or via group membership

Issue #80, opened 2025-02-28
Permissions based on groups not always calculated correctly in case of multiple group memberships


gdiepen:
When a user is a member of multiple groups, the logic for checking membership is not always done in the correct way.

Suppose you have a user that is a member of team1 and team2 and the user has no direct user permissions on any experiment.

If the team2 does not have any permissions on the given experiment, it will still assign None to the user_perms variable in the `get_user_groups_experiement_permissions` function, which will result in an exception in the calling function, which will result in the default fallback being used, which in my case would NO_PERMISSIONS.

One of the underlying problems might be the priority of 100 on the NO_PERMISSIONS permission. I would say that if a user is in two groups team1 and team2 and team1 has READ access on Experiment1 and team2 has NO_PERMISSIONS access on Experiment1, the result should be that this user has access (based on the team1 membership). This might be connected with assuming a default of MANAGE for DEFAULT_FLOW_PERMISSION.

If we start with the DEFAULT permission to be NO_PERMISSIONS, we can utilize role based access to experiments / models based on memberships (where the highest permission wins)

I would propose the following changes:
* Change the default AppConfig.DEFAULT_MLFLOW_PERMISSION to "NO_PERMISSION"
* Change the NO_PERMISSIONS priority level from 100 to 0
* Change `get_permission_from_store_or_default` function to basically create a list of all user and group permissions. If this list is empty -> fallback to the default permission. If the list is not empty, return `max( default_permission, max(all_user_and_group_permissions) )`. Using this first check with empty list ensures we can log debug information that we falled back to the default because nothing was explicitly defined for this user

Let me know what you think about this updated logic. Same as with my other issues, happy to create PRs for it, since I am anyway working on trying to get it to work here.

gdiepen:
Quick update on this one, I have a basic version of this implemented and working. Will need to do a bit of cleaning up the code as well as create/modify unit tests and can create a PR out of this

kharkevich:
hi @gdiepen could you please check 3.4.1rc1 from test pypi:  https://test.pypi.org/project/mlflow-oidc-auth/ 
it should fix the issue

gdiepen:
Where is the source for that release?

I am actually using the latest main branch in my own local environment and then apply all my required changes

gdiepen:
Found it already, will check Monday if this works or not

gdiepen:
I went through it and it seems to work, though I had already implemented a similar logic inside of my PR #86 

[...]

PR #86, opened 2025-03-10
feat: rewrite permission logic to no permissions having lowest priority

gdiepen:
Instead of giving `NO_PERMISSIONS` a higher priority than all the other priorities, which allows for weird blocking of access if you have both MANAGE PERMISSIONS as well as NO_PERMISSIONS, this PR will modify the permission logic to be giving you the access of the highest allowed permissions based on the default permission, your user permissions, as well as each of your groups permissions.

In order to check what the type is that gave you access, we start with type fallback, then if the user permission prioerity is strictly higher than the fallback one, we change the type to user and if after that any of the group permissions have a strictly higher priority, we consider the group to be the type that granted this access.

For example, if the default permission is READ and your user also has explicit READ permissions, we consider the type to still be fallback (the fact that the user has read access does not grant anything additional above the defaults).

By using this logic and assigning NO_PERMISSIONS the lowest priority, you automatically solve the problem that if you are a member of two groups and one group gives you MANAGE access and the other one NO PERMISSIONS, you end up with no permissions in the current state (in the state with this pr, you would end up with MANAGE).

Because of the changes in this PR, this also Implements #80

This also might address #83, though without the use of an extra config variable

### DIFF
--- a/mlflow_oidc_auth/permissions.py
+++ b/mlflow_oidc_auth/permissions.py
@@ -43,7 +43,7 @@ MANAGE = Permission(
 
 NO_PERMISSIONS = Permission(
     name="NO_PERMISSIONS",
-    priority=100,
+    priority=0,
     can_read=False,
     can_update=False,
     can_delete=False,
--- a/mlflow_oidc_auth/permissions.py
+++ b/mlflow_oidc_auth/permissions.py
@@ -70,7 +70,7 @@ def _validate_permission(permission: str):
         )
 
 
-def compare_permissions(permission1: str, permission2: str) -> bool:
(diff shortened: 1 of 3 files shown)


kharkevich:
Thanks @gdiepen — this points at a genuine, still-open bug (#80): with `NO_PERMISSIONS` at the highest priority, a `NO_PERMISSIONS` group can override a `MANAGE` group. Closing because the `legacy` base branch is being decommissioned and this diff targets the pre-FastAPI code. The underlying issue #80 stays open and will need a fresh `main`-targeted fix (plus a deliberate decision on whether NO_PERMISSIONS means explicit-deny or absence). Not a won't-fix — a wrong-target retirement.

Issue #90, opened 2025-03-17
Running MlFlow behind reverse-proxy does not work with the plugin


hahahannes:
Hi!

# General Use Case
I am trying to run MlFlow behind a reverse proxy which routes traffic based on the path. I also want to use the plugin for authentication and authorization.

# Environment
# Steps to reproduce a Bug
1. Run a reverse proxy e.g. nginx via `docker compose up`
```yaml
services:
  nginx:
    image: nginx:latest
    volumes:
      - ./nginx.conf:/etc/nginx/nginx.conf
    network_mode: host
```
`nginx.conf`:
```
events {}

http {
    # include the default mime.types to map file extensions to MIME types
    include /etc/nginx/mime.types;

    server {
        # set the root directory for the server (we need to copy our
        # application files here)
        root /usr/share/nginx/html;

        # set the default index file for the server (Angular generates the
        # index.html file for us and it will be in the above directory)
        index index.html;
        	
        location /models/ {
            rewrite ^/models(.*)$ $1 break;
            proxy_pass http://localhost:8080;
        }

    }
}

```
`.env`:
```
OAUTHLIB_INSECURE_TRANSPORT=1
OIDC_REDIRECT_URI=http://localhost/models/callback
OIDC_DISCOVERY_URL=URL
OIDC_CLIENT_ID=ID
OIDC_CLIENT_SECRET=SECRET
LOG_LEVEL=DEBUG
OIDC_SCOPE=openid email
```
2. Create the `.env` file and run the dev server `scripts/run-dev-server.sh`
3. Open `http://localhost/models`

# Expected Result
I would expect that I get redirect to the Identity Provider to then authenticate myself and to then get redirected to `http://localhost/models/callback` to finish the login.

# Actual Result
You will get redirected to `http://localhost/login` which does not exist.
This is caused by a redirect at https://github.com/mlflow-oidc/mlflow-oidc-auth/blob/main/mlflow_oidc_auth/hooks/before_request.py#L200 and https://github.com/mlflow-oidc/mlflow-oidc-auth/blob/main/mlflow_oidc_auth/templates/auth.html#L10C1-L11C1. Depending on if automatic redirect is enabled.

# Possible Fix
I have a working prototype which works when running behind a reverse proxy and also when not running behind a reverse proxy.

## URL generation in the backend
[...]

PR #91, opened 2025-03-17
feat: support running behind reverse proxy with path based routing

hahahannes:
Fix #https://github.com/mlflow-oidc/mlflow-oidc-auth/issues/90

# Feature
This feature will enable to run MlFlow together with this plugin to run behind a reverse proxy that routes traffic based on the path, e.g. `/models`. This requires some changes in the backend and frontend to use the prefix path. Ideally this prefix path should not be hardcoded or configured.

## URL generation in the backend
We can use Werkzeugs `ProxyFix` (https://werkzeug.palletsprojects.com/en/stable/middleware/proxy_fix/) to use `X-Forwarded-Prefix` from the reverse proxy. This will enable `url_for` to generate the correct URL with the prefix

## URL generation in the frontend
The UI needs to be configurable to use a base prefix. This could is done via en environment setting which per default is empty. 

# How to use the feature
The user has to properly configure their reverse proxy as well as override the base prefix in the UI code.
 
## Required changes at Reverse Proxy
In order to work one needs to adjust his reverse proxy to set the needed header. Also at the moment accessing `/models` does not work as the requests for the static mlflow files as well as for the current username (https://github.com/mlflow-oidc/mlflow-oidc-auth/blob/main/mlflow_oidc_auth/hack/menu.html#L42) will ignore the path. I did not want to introduce a new config variable to add the base path here manually so the quick fix is to redirect requests to `/models` to `/models/`. Then the relative paths will work again.
```
proxy_set_header X-Forwarded-Prefix /models;
location /models {
    return 301 /models/;
}
```

## Required changes during build time
As the UI is part of the package, there is no easy way to set the `basePrefix` in the angular configuration without hardcoding it. Therefore, I am overriding it before building the UI. I have not find a way yet to set it dynamically. I tried overriding the compiled javascript in the pip package but was not successful as the production js code gets optimized 
```dockerfile
RUN git clone https://github.com/mlflow-oidc/mlflow-oidc-auth
RUN find '.' -name '*.ts' -exec sed -i -e 's@basePrefix: ""@basePrefix:"/models"@g' {} \;
RUN yarn install
RUN yarn release 
RUN python -m pip install .
```

### DIFF
--- a/mlflow_oidc_auth/app.py
+++ b/mlflow_oidc_auth/app.py
@@ -1,5 +1,7 @@
 import os
 
+from werkzeug.middleware.proxy_fix import ProxyFix
+
 from mlflow.server import app
 from flask_session import Session
 
@@ -86,6 +88,8 @@ app.add_url_rule(rule=routes.UPDATE_GROUP_MODEL_PERMISSION, methods=["PATCH"], v
 # Add new hooks
 app.before_request(before_request_hook)
 app.after_request(after_request_hook)
+
+app.wsgi_app = ProxyFix(app.wsgi_app, x_prefix=1)
 
 # Set up session
 Session(app)
--- a/mlflow_oidc_auth/hack/menu.html
+++ b/mlflow_oidc_auth/hack/menu.html
@@ -45,8 +45,8 @@
                 return;
             }
             var username = response.display_name;
-            var oidcLink = "/oidc/ui/";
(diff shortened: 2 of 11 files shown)


kharkevich:
hi @hahahannes,  thanks for the wonderful idea! I reworked it a little and implemented it in #136

PR #99, opened 2025-04-09
fix: filter out EntraId groups without displayName or with duplicate displayName

Eleirbag89:
## Summary

This pull request resolves an issue experienced during the EntraId setup in our company. The server was returning a 500 error because the plugin provided groups with a `displayName` value of `None`, causing a violation of database constraints when saving new group entries.

## Problem Details

The error originates from an offending group entry, which appears to be related to the tenant's Global Administrator role. An example of the problematic entry is shown below:

```json
{
    "@odata.type": "#microsoft.graph.directoryRole",
    "id": "XXXXXXX",
    "deletedDateTime": "None",
    "description": "None",
    "displayName": "None",
    "roleTemplateId": "None"
}
```
## Fix
The implemented fix introduces a simple validation check that filters out any groups where the displayName is not set. This prevents the insertion of problematic entries into the database and avoids the server crash. The solution has been fully tested on our servers.

Please review the changes and let me know if you have any questions.


kharkevich:
hi @Eleirbag89,

Thanks for your contribution! Could you please fix the formatting and unit test issues?

PR #121, opened 2025-06-19
feat: support pagination for Entra ID group

mis-association:
2 issue are identified when we test oidc-auth :
Issue when user from EntraID has a lot of group, we need browse all page with '@odata.nextLink'

Issue when user in EntraID with tech account from Onpremise AD, email is also retrieve from  "prefered_username" when email field is not setup from TechAccount in AD

This PR is to resolve them.

We test all use case locally with real ad , tenant etc and user and is okay


mis-association:
Hello @kharkevich 
Thanks for the review and feedback.

I just added a test.

I'm open to any refactoring or other adjustments.

mis-association:
Hello @kharkevich, 
Do you thinks there is any action pending on my side to complete this PR ?

PR #137, opened 2025-08-01
feat: add fortawesome css to package

cagdasbas:
Because the fortawesome css is not included in the built package, web ui fetches it from the internet which is not possible for airgap installations. This change request adds the css to package itself so even with no internet access, the web ui works properly.

### DIFF
--- a/web-ui/angular.json
+++ b/web-ui/angular.json
@@ -24,7 +24,7 @@
             "tsConfig": "tsconfig.app.json",
             "inlineStyleLanguage": "scss",
             "assets": ["src/favicon.ico", "src/assets"],
-            "styles": ["./node_modules/@angular/material/prebuilt-themes/indigo-pink.css", "src/styles.scss"],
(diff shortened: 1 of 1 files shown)


kharkevich:
hi @cagdasbas,
could you please provide an example when it is loaded from outside?

cagdasbas:
Hi @kharkevich,
Before this patch, the web UI fetched the Font Awesome CSS from Cloudflare:
image

This causes the web UI to hang if the server has no internet access. Every page navigation triggers a new fetch attempt, leading to repeated delays.

This patch bundles the CSS into the final wheel, so it's served directly from the deployed MLflow server instead of the internet:
image

I tested this by building the wheel with the existing build script and installing it in my Dockerfile. I can confirm the server now runs without any hangs.
Thanks!

cagdasbas:
Hi @kharkevich ,

It turns out, because our mlflow version is 2.31, oidc plugin version 3.7 is installed which has this issue. Thanks for the tip! Please feel free to close the PR.

PR #148, opened 2025-09-26
fix: LoggedModels do not show up when using Group permissions

batuhanozoge:
We have a use-case where we do not issue any user rights and do authorization solely based on group permissions. (Our fallback permission is NO_PERMISSIONS so users all use the fallback). In such case the after_request hook filters out all the LoggedModels since it fetches all experiments based on user rights "store.list_experiment_permissions(username)"

I believe this should take into account the effective permissions of the user(including group) and not just user permissions. That is why I have modified it to use effective_experiment_permission(model.experiment_id, username).permission.can_read  

I haven't adjusted the tests yet they assert store.list_experiment_permissions(username) to be called once.

### DIFF
--- a/mlflow_oidc_auth/utils/data_fetching.py
+++ b/mlflow_oidc_auth/utils/data_fetching.py
@@ -22,7 +22,7 @@ from mlflow.store.entities.paged_list import PagedList
 from mlflow_oidc_auth.config import config
 from mlflow_oidc_auth.permissions import get_permission
 from mlflow_oidc_auth.store import store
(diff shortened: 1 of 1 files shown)


kharkevich:
Landed via #266 (merged) — thanks @batuhanozoge. Reimplemented on the current `data_fetching.py`: logged-model read filtering now uses `effective_experiment_permission`, so models in group-granted experiments show up even when `DEFAULT_MLFLOW_PERMISSION=NO_PERMISSIONS`. Your fix is on `main`.

PR #157, opened 2025-11-18
feat: enable pkce

ivanmanan:
feat: Currently, PKCE is not enabled by default. Set PKCE via the oauth2/client.py parameter - https://github.com/authlib/authlib/blob/main/authlib/oauth2/client.py


kharkevich:
Landed via #266 (merged) — thanks @ivanmanan. PKCE is now available via `OIDC_CODE_CHALLENGE` (e.g. `S256`), reimplemented on the current `ensure_oidc_client_registered` client_kwargs. On `main`.

PR #158, opened 2025-11-22
fix: restrict user and group visibility for non-admin users

hsteude:
Hi there,

I noticed that non-admin users can see all usernames and groups in the management UI. I'm wondering if this was intentional, because the admin checks for `list_users()` were already implemented but commented out.

  For our use case, regular users shouldn't be able to see other users' information in the UI.

  ## Changes
  - Re-enabled admin checks in `list_users()` (were commented out)
  - Added admin checks to `list_groups()` (following `list_prompts()` pattern)
  - Non-admin users now only see themselves and their own groups

 What do you think?

 BTW: Thanks, this is an awesome project!!


kharkevich:
Thanks @hsteude — the concern is real and still valid: non-admins can enumerate all users/groups on `main`. Closing because the `legacy` base branch is being decommissioned (the codebase moved to the FastAPI architecture, so this diff no longer applies). The visibility restriction will be reimplemented against the current routers as its own `main`-targeted change. Not a won't-fix — a wrong-target retirement.

PR #166, opened 2025-12-02
chore: configure Dependabot security updates

felipeolifre:
Given this is a public project, I suggest setting up Dependabot security updates: https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/about-dependabot-security-updates#about-dependabot-security-updates


felipeolifre:
Actually, it seems the security updates are already enabled for this repo and do not require to have a `dependabot.yml` file in place.

PR #169, opened 2025-12-18
feat: allow cookie session type

OstlerDev:
When attempting to run this plugin load balanced between multiple servers I ran into an issue with the default config. I did not want to setup a shared Redis server, so I have added support for Cookie based auth.

Before this change `mlflow_oidc_auth/config.py` would always try to import a session backend module for whatever SESSION_TYPE was set to which would cause a crash because there was no file named `mlflow_oidc_auth/session/cookie.py` when it attempted to dynamically load the Session Backend provider. 

This fix removes the crash from happening, and allows Cookie auth to continue.

---

## AI Summary
Fix intermittent OIDC login failures in multi-replica deployments by making cookie sessions an opt-in stateless session backend.

## Problem
When MLflow is deployed with multiple replicas behind a load balancer, `/login` and `/callback` may be served by different pods. The plugin stores the OIDC CSRF state in the Flask session (`session["oauth_state"]`), but the default session backend is server-side filesystem (`SESSION_TYPE=cachelib`), which is local to a single pod. As a result, `/callback` on a different pod can’t find the stored state and login intermittently fails.

## Root cause
- `mlflow_oidc_auth/app.py` always initialized Flask-Session (`Session(app)`), enabling server-side sessions.
- The default `SESSION_TYPE=cachelib` stores session state on the pod filesystem (`/tmp/flask_session`), which is not shared across replicas.
- A shared `SECRET_KEY` alone does not fix this because the state is not in the cookie; it is stored server-side.

## Solution
- Add an opt-in stateless session mode: `SESSION_TYPE=cookie`.
- Make Flask-Session initialization conditional:
  - For `SESSION_TYPE=cookie`, do not initialize Flask-Session (use Flask’s signed cookie sessions).
  - For server-side backends (e.g. `cachelib`, `redis`), keep initializing Flask-Session.
- Keep backward compatibility:
  - Default remains `SESSION_TYPE=cachelib`.
  - `SESSION_TYPE=redis` remains supported.
- Update documentation in the existing format to include the new cookie backend as an additional option.

## Configuration guidance
- Single instance (default behavior):
  - `SESSION_TYPE=cachelib` (default)
- Multi-instance behind a load balancer without sticky sessions (no Redis required):
  - `SESSION_TYPE=cookie`
  - Set a consistent `SECRET_KEY` across all instances
- Multi-instance with shared server-side session store:
  - `SESSION_TYPE=redis` (+ Redis configuration)

## Tests
- Added unit tests to verify:
  - Flask-Session is not initialized when `SESSION_TYPE=cookie`.
  - The `/login` → `/callback` state flow succeeds using Flask’s signed cookie sessions across requests.
- Test suite: `pytest` passes (`393 passed`).

## Security notes
- Cookie sessions are signed (integrity-protected). The plugin stores only small session state required for login flow (e.g. `oauth_state`, `username`) and does not store OIDC access tokens in cookies.
- For cookie sessions across replicas, `SECRET_KEY` must be identical across all instances.

## Files changed
- `mlflow_oidc_auth/app.py`: conditional Flask-Session initialization
- `mlflow_oidc_auth/config.py`: cookie session opt-in support without importing a session module
- `docs/configuration.md`: add “Session Backend: Cookie” section after existing backends
- `mlflow_oidc_auth/tests/test_session_modes.py`: new tests for cookie mode and conditional init

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -72,6 +72,11 @@ The application can be configured through environment variables, dotenv files, o
 | `REDIS_SSL` | Boolean | `false` | Use SSL connection to Redis |
 | `REDIS_USERNAME` | String | None | Redis username for authentication |
 
+### Session Backend: Cookie
+*Used when `SESSION_TYPE=cookie`*
+
+This backend is intended for running multiple instances behind a load balancer (no sticky sessions). You must set a consistent `SECRET_KEY` across all instances.
+
 ### Cache Configuration
 
 | Variable | Type | Default | Description |
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -77,6 +77,10 @@ The application can be configured through environment variables, dotenv files, o
 
 This backend is intended for running multiple instances behind a load balancer (no sticky sessions). You must set a consistent `SECRET_KEY` across all instances.
 
+| Variable | Type | Default | Description |
+|----------|------|---------|-------------|
+| `SECRET_KEY` ⚠️ | String | None | Application secret key used to sign and encrypt session cookies. Must be set and identical across all instances. |
+| *(none)* | - | - | No additional cookie-backend-specific variables are required beyond the core session configuration. |
 ### Cache Configuration
 
 | Variable | Type | Default | Description |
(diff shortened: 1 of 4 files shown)


kharkevich:
Thanks @OstlerDev. Closing as obsolete: `main` no longer uses Flask-Session/cachelib — it now uses Starlette signed-cookie sessions by default, which is exactly the stateless, multi-replica-safe behavior this PR added. The crash it fixed can no longer occur.

Issue #198, opened 2026-02-04
React UI API calls does not respect base path


Korel:
## Description
When running mlflow-oidc behind a reverse proxy with a path prefix (/mlflow in my case), the React UI does not use the prefix for the API calls. GET requests work correctly. I believe this is a regression from Angular UI.

## Environment

  - mlflow-oidc-auth: v6.6.4
  - MLflow: 3.9.0
  - Reverse proxy: nginx ingress -> Istio ingressgateway -> VirtualService with X-Forwarded-Prefix header configured to /mlflow

## Steps to reproduce

1. Deploy MLflow behind a reverse proxy at a subpath (e.g., /mlflow)
2. Configure X-Forwarded-Prefix: /mlflow header
3. Verify /mlflow/oidc/ui/config.json returns {"basePath": "/mlflow", ...}
```
❯ curl -fsSL https://<domain>/mlflow/oidc/ui/config.json | jq
{
  "basePath": "/mlflow",
  "uiPath": "/mlflow/oidc/ui",
  "provider": "Sign in with Keycloak",
  "authenticated": false
}
```
4. Navigate to the OIDC UI and try to create an access token

## Expected result

PATCH request goes to /mlflow/api/2.0/mlflow/users/access-token

## Actual result

PATCH request goes to /api/2.0/mlflow/users/access-token (missing /mlflow prefix), resulting in a 404.

PR #199, opened 2026-02-04
fix: prefix relative URLs with runtime config base path

Korel:
## Description (Updated after Copilot review):
This PR fixes issue #198 by ensuring all API calls and navigation properly handle the basePath from runtime config.

## Changes

**1. BasePath-aware API calls**
Updated services and components to use `resolveUrl` for proper basePath prefixing:
- Access token modal
- Trash, user, and webhook services
- Permissions management (normal and regex views)
- Added tests verifying basePath prefixing from runtime config

**2. Fixed double base path navigation bug**
Discovered and fixed an additional bug where row action URLs were being double-prefixed:
- Changed row action URLs to use router-relative paths
- Ensures `uiPath` is only applied once by the router
- Added basename-aware router tests to validate correct navigation

## Testing

* Added test cases for the code changes and additions.
* Deployed behind proxy with custom basePath (/mlflow) by building the UI through the following dockerfile:
```dockerfile
FROM ghcr.io/mlflow-oidc/mlflow-tracking-server:3.9.0-6.6.4-20260130
COPY mlflow_oidc_auth/ui/ /mlflow/.venv/lib/python3.12/site-packages/mlflow_oidc_auth/ui/

### DIFF
--- a/web-react/src/core/components/access-token-modal.tsx
+++ b/web-react/src/core/components/access-token-modal.tsx
@@ -1,7 +1,6 @@
 import React, { useState, useCallback, useRef } from "react";
 import { useUser } from "../hooks/use-user";
-import { http } from "../services/http";
-import { resolveUrl } from "../services/api-utils";
+import { request } from "../services/api-utils";
 import { STATIC_API_ENDPOINTS } from "../configs/api-endpoints";
 import { faCopy } from "@fortawesome/free-solid-svg-icons";
 import { Button } from "../../shared/components/button";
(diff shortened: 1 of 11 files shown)


Korel:
> Thanks for the comprehensive fix! The updates look great, and I’ve verified that this resolves the navigation problems.
> 
> While reviewing the changes, I noticed a consistent pattern where `resolveUrl` is followed immediately by a call to `http`. Since this PR establishes and validates the `resolveUrl` approach, I’d like to request a small refactor as part of this PR to centralize this logic.
> 
> Specifically, could you please introduce a helper (for example, `request`) in `api-utils.ts` that encapsulates URL resolution, basePath handling, and the `http` call in one place?
> 
> Smth like
> 
> ```
> export async function request<T>(
>   endpoint: string,
>   options: RequestOptions & { queryParams?: QueryParams } = {},
> ): Promise<T> {
>   const { queryParams, ...httpOptions } = options;
>   const url = await resolveUrl(endpoint, queryParams || {}, options.signal);
>   return http<T>(url, httpOptions);
> }
> ```
> 
> I think this would apply cleanly to `access-token-modal.tsx`, ` use-permissions-management.ts`, `webhook-service.ts`, `trash-service.ts`, `user-service.ts`, and `create-api-fetcher.ts` as well, and it would help keep the API layer more consistent going forward.

Thanks for the review! I have implemented the requested refactor. Please let me know if additional changes are needed.

PR #203, opened 2026-02-11
fix: uri-encode group names

TheChristophe:
In our deployment (rel.: #187), we determine groups based on a roles that contain '#', which is a reserved URL characer (anchor). The frontend inserts group names directly into the frontend/API URLs, causing it to fail. This fix ensures they are encoded first.


TheChristophe:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/mlflow-oidc/mlflow-oidc-auth/pull/203#pullrequestreview-3782939747)

PR #215, opened 2026-02-24
feat: make username field configurable

TheChristophe:
Make username and display name source configurable

Various service provides may be against using particular OIDC fields like emails as unique identifier, as they may change and end up creating two accounts in MLFlow, rendering the old one inaccessible

Similarly, in various research environments, using emails as display name may be preferrable so users can contact each others easier


reedswenson-se:
This sure would be useful with our IDP.  They have fixed claims and `name` isn't one of them.  

Any idea when this can be merged and released?

kharkevich (approved):

reedswenson-se:
Thanks so much!

PR #225, opened 2026-03-18
feat: add the option disable TLS verification on OIDC provider

stergem:
Optionally disable TLS verification on the OIDC provider. Useful for local development with self-signed certs


kharkevich:
Landed via #266 (merged) — thanks @stergem. `OIDC_VERIFY_SSL` (default true) now toggles TLS verification on discovery/JWKS/token requests, reimplemented on the current `oauth.py`/`auth.py`/`config.py`. On `main`.

PR #243, opened 2026-04-29
fix: improve auth-failure diagnostics

remidebette:
When token validation fails, the bearer/basic auth handlers log only the exception class name with no message, traceback, URL, or token claim. Operators investigating intermittent 401s cannot tell apart token expiry, JWKS network errors, signature mismatch, or claim issues — all collapse to the same opaque "Invalid token" log line.

<del>JWKS fetch in `_get_oidc_jwks` calls `requests.get` without a timeout. A slow or hung IdP can block request threads until the OS-level TCP timeout (~2 minutes), turning a brief upstream blip into a wave of authentication failures.<del>

Changes:
- `auth_middleware.py`: log exception message and traceback on basic/bearer auth failure, while keeping the returned error string stable so existing tests and callers are unaffected.
- <del>`auth.py`: add a 10s timeout on the OIDC discovery and JWKS HTTP calls.<del>

### DIFF
--- a/mlflow_oidc_auth/middleware/auth_middleware.py
+++ b/mlflow_oidc_auth/middleware/auth_middleware.py
@@ -83,7 +83,7 @@ class AuthMiddleware(BaseHTTPMiddleware):
             else:
                 return False, None, "Invalid basic auth credentials"
         except Exception as e:
-            logger.error("Basic auth error: %s", type(e).__name__)
+            logger.error("Basic auth error: %s: %s", type(e).__name__, e, exc_info=True)
             return False, None, "Invalid basic auth format"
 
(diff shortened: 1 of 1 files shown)


kharkevich:
hi @remidebette,
can you please make timeout option configurable and move defaults to configuration,
also, please make sure that all test adjusted (or created new one) to cover with functionality w/o issues

remidebette:
I moved the Timeout changes to its own PR, https://github.com/mlflow-oidc/mlflow-oidc-auth/pull/246
Only the log changes that allowed investigation is kept here

kharkevich (approved):

Issue #195, opened 2026-01-31
feat: Add configurable permissions to restrict experiment and model creation


anvaari:
Hi, I have this requirement but create this Issue using DeepWiki, JFYI :)

## Problem Description
Currently, the MLflow OIDC Auth system allows all authenticated users to create experiments and registered models without any restrictions. This "create-first, control-after" approach can lead to resource sprawl in enterprise environments where administrators need tighter control over MLflow resources.

The creation endpoints (`CreateExperiment` and `CreateRegisteredModel`) bypass permission checks entirely, automatically granting `MANAGE` permissions to creators.

## Use Cases
- **Enterprise Governance**: Organizations need to control who can create experiments/models to maintain resource organization and cost control
- **Multi-tenant Environments**: Different teams or departments may have different creation privileges
- **Compliance Requirements**: Some regulated environments require strict resource creation controls

## Proposed Solution
Add a new configuration option `RESTRICT_RESOURCE_CREATION` (default: false) that enables permission checks on creation endpoints. When enabled:

1. **Configuration**: Add new config variables:
   ```bash
   RESTRICT_RESOURCE_CREATION=true
   DEFAULT_CREATION_PERMISSION=MANAGE  # or EDIT, READ
   ```

2. **Permission Enforcement**: Extend `before_request_hook` to validate creation permissions:
   - Check `CREATE_EXPERIMENT` permission for experiment creation
   - Check `CREATE_MODEL` permission for registered model creation
   - Return 403 Forbidden if user lacks required permission

3. **Permission Levels**: Introduce new permission levels:
   - `CREATE_EXPERIMENT`: Allow experiment creation
   - `CREATE_MODEL`: Allow registered model creation
   - `CREATE_ALL`: Allow both (for admin users)

### 1. Configuration Extension
Add to `config.py` :
```python
RESTRICT_RESOURCE_CREATION: bool = False
DEFAULT_CREATION_PERMISSION: str = "MANAGE"
```

### 2. Permission Schema Updates
Extend database schema to include creation permissions:
- Add `can_create_experiments` and `can_create_models` columns to user permissions tables
- Add corresponding group permission columns

### 3. Hook Modifications
Update `before_request_hook` to include:
```python
if config.RESTRICT_RESOURCE_CREATION:
    if endpoint == "CreateExperiment":
        require_permission(user, "CREATE_EXPERIMENT")
    elif endpoint == "CreateRegisteredModel":
[...]

PR #247, opened 2026-05-04
feat: enforce authorization on CreateExperiment and CreateRegisteredModel

### LINKED ISSUE #202: DEFAULT_MLFLOW_PERMISSION -> NO_PERMISSIONS takes no effect
Hi,

Users are still able to create experiments even tho the default permission is none and their group they are part of has no assigned permission.

remidebette:
## Summary

Resurrects @nitaibezerra's original work from #195/#202, rebased onto current `main` and made workspace-aware.

Currently any authenticated user can create experiments and registered models regardless of `DEFAULT_MLFLOW_PERMISSION` or group/regex permissions — there's no before-request validator on `CreateExperiment`/`CreateRegisteredModel`, only an after-request handler that grants `MANAGE` to the creator after the resource already exists.

This PR adds opt-in creation enforcement gated by a new `RESTRICT_RESOURCE_CREATION` config flag (default: `false`), so existing deployments see no behaviour change unless they enable it.

Fixes #202.

## Changes

Three rebased commits from @nitaibezerra plus one workspace-correctness fix:

1. **`fix: enforce authorization on CreateExperiment and CreateRegisteredModel`** — adds `validate_can_create_experiment` and `validate_can_create_registered_model` validators wired into `BEFORE_REQUEST_HANDLERS`. Resolves permission by **resource name** (the resource ID doesn't exist yet at creation time) using regex/group-regex sources only.
2. **`feat: add RESTRICT_RESOURCE_CREATION flag for backward compatibility`** — gates the new validators behind an opt-in config flag (default `false`).
3. **`docs: add resource creation authorization documentation`** — documents the flag and the resolution order.
4. **`fix(auth): respect workspace permissions for resource creation`** *(new in this rebase)* — extracts the existing workspace-fallback logic from `resolve_permission` into a small helper and applies it on the creation paths. Without this, with `MLFLOW_ENABLE_WORKSPACES=true` a user with no workspace permission would fall through to `DEFAULT_MLFLOW_PERMISSION` and could still create resources.

## Behaviour

When `RESTRICT_RESOURCE_CREATION=false` (default): no change — anyone authenticated can create.

When `RESTRICT_RESOURCE_CREATION=true`:

| Scenario | Result |
|---|---|
| User has matching regex / group-regex granting EDIT+ on the requested name | Allow |
| Workspaces disabled, no regex match | Falls back to `DEFAULT_MLFLOW_PERMISSION` |
| Workspaces enabled, no regex match, user has EDIT+ on request workspace | Allow |
| Workspaces enabled, no regex match, no workspace permission | **Deny** |

## Test plan

- [x] `pytest mlflow_oidc_auth/tests/validators/test_experiment.py mlflow_oidc_auth/tests/validators/test_registered_model.py mlflow_oidc_auth/tests/hooks/test_before_request.py mlflow_oidc_auth/tests/utils/test_permissions.py` — 158/158 passing locally
- [x] New tests cover: regex match, regex miss with workspaces enabled and a workspace permission, regex miss with workspaces enabled and no workspace permission (deny), regex miss with workspaces disabled
- [ ] Maintainer review

## Notes

@nitaibezerra is preserved as the original author on commits 1-3. Happy to squash on merge if maintainers prefer a single-commit PR — let me know.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -31,6 +31,7 @@ The application is configured through environment variables, `.env` files, or pl
 |----------|------|---------|-------------|
 | `DEFAULT_MLFLOW_PERMISSION` | String | `MANAGE` | Default permission level when no explicit permission is found. Options: `READ`, `USE`, `EDIT`, `MANAGE`, `NO_PERMISSIONS`. See [Permissions](permissions) |
 | `PERMISSION_SOURCE_ORDER` | String | `user,group,regex,group-regex` | Comma-separated order for evaluating permission sources. The first source with a matching permission wins. See [Permissions](permissions#permission-source-order) |
+| `RESTRICT_RESOURCE_CREATION` | Boolean | `false` | When enabled, enforce permission checks on experiment and registered model creation. See [Resource Creation Authorization](permissions#resource-creation-authorization) |
 
 ### Database
 
--- a/docs/permissions.md
+++ b/docs/permissions.md
@@ -121,6 +121,41 @@ When a user creates a resource, the plugin automatically grants them `MANAGE` pe
 - `CreateGatewayModelDefinition` → MANAGE on the new model definition
 - `CreateWorkspace` → MANAGE workspace permission (when workspaces enabled)
 
+## Resource Creation Authorization
+
+By default, any authenticated user can create experiments and registered models. This matches upstream MLflow behavior.
+
+When `RESTRICT_RESOURCE_CREATION` is set to `true`, the system enforces permission checks on `CreateExperiment` and `CreateRegisteredModel` requests. The user must have at least **EDIT** permission (`can_update`) to create a resource.
+
+### How creation permissions are resolved
+
+Since the resource does not exist yet at creation time, only **name-based** permission sources apply:
+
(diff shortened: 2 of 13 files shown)


remidebette:
I just noticed #245 after rebasing

The spirit behind this PR and the past one was that - **pre-workspace** - any experiment creation was allowed (as a bug), even the ones which subsequent access was being forbidden by a regex. 
For example in my case of project based prefix `<project>/<exp name>` some people would be allowed some project prefix but not all.
So that if users did typos in the project name, they ended up creating experiments they would not be able to access.

This gets restricted by this PR, in the sense that a user needs **experiment** EDIT access to be able to create such an exp.

The issue is that with #245 and **post-workspace**, the responsibility of creating exps is not in the exps access rights anymore but in the **workspace** access rights (and should be Workspace MANAGE as per https://mlflow-oidc.github.io/mlflow-oidc-auth/#/workspaces?id=workspace-permissions

I was trying to see how to adapt the current non-workspace code for workspaces, but it seems irrelevant now.

Should we still plan on merging for the bug in: `MLFLOW_ENABLE_WORKSPACES=false` @kharkevich, or would you rather encourage users to switch to workspaces instead ?

PR #249, opened 2026-05-06
feat(auth): re-authenticate sessions against IdP token expiry

### LINKED ISSUE #242: Users are never re-authenticated against the IdP after initial OIDC login
## Observation

Once a user completes the OIDC login flow, they remain authenticated to the
MLflow server for ~14 days regardless of any IdP-side events (account
deactivation, group membership change, MFA enrollment, password reset, etc.).
There is no mechanism that re-validates the session against the IdP.

## Why this happens

Looking at the current code on `main`:

1. After a successful callback, only the username is persisted to the session
   ([`mlflow_oidc_auth/routers/auth.py`](https://github.com/mlflow-oidc/mlflow-oidc-auth/blob/main/mlflow_oidc_auth/routers/auth.py)):

   ```python
   session["username"] = email
   session["authenticated"] = True
   ```

   No access token, no `exp` claim, no IdP-issued expiry timestamp is stored.

2. `AuthMiddleware._authenticate_session`
   ([`mlflow_oidc_auth/middleware/auth_middleware.py`](https://github.com/mlflow-oidc/mlflow-oidc-auth/blob/main/mlflow_oidc_auth/middleware/auth_middleware.py))
   only checks for the presence of `session["username"]`. No expiry check, no
   IdP re-validation, no token refresh.

3. `StarletteSessionMiddleware` is registered without `max_age`
   ([`mlflow_oidc_auth/app.py`](https://github.com/mlflow-oidc/mlflow-oidc-auth/blob/main/mlflow_oidc_auth/app.py)),
   so the cookie falls back to Starlette's default of 14 days.

The combination means the only thing that ends a user's session is the
14-day cookie TTL or an explicit sign-out.

## Why it matters

For deployments that rely on the IdP as the source of truth for who is allowed
in (deprovisioning users, rotating group membership, enforcing MFA), this is
effectively a stale-auth window of up to two weeks. It also means
`OIDC_GROUP_NAME` / `OIDC_ADMIN_GROUP_NAME` changes (and any plugin-driven
group resolution) only take effect on the next login, which can be far in the
future for an active user.

## Possible fixes

There are several layers, each with different trade-offs — would love
maintainer input on which direction is acceptable upstream:

1. **Hard session cap.** Pass `max_age=<seconds>` to `StarletteSessionMiddleware`,
   exposed via a new config (e.g. `SESSION_MAX_AGE_SECONDS`). One-liner;
   forces a full OIDC redirect flow at the cap.

2. **Respect the OIDC token expiry.** Store the access/ID token `exp` claim
   in the session at callback time and reject the session in
   `_authenticate_session` once expired. Aligns plugin behavior with the
   IdP's token lifetime instead of an arbitrary cookie TTL.

3. **Silent refresh with refresh tokens.** Request `offline_access`, persist
   the refresh token (server-side, given the security implications of
   storing refresh tokens in cookies), and refresh on expiry. This re-validates
   with the IdP — and re-pulls group memberships — without forcing visible
   logins. Largest change but the most correct.

4. **Force IdP-side re-auth.** Pass `prompt=login` or `max_age=<seconds>` on
   the `authorize_redirect` call so the IdP enforces re-auth on each
   redirect. Complementary to (1)/(2).

A reasonable minimum seems to be **(1) + (2) together**: a hard cap plus
respecting the IdP's `exp`. (3) feels like the right long-term direction.

## Questions for maintainers

- Is this behavior intentional / known?
- Which of the above directions would you accept as a PR? In particular,
  is storing the access token's `exp` in the session acceptable, or would
  you prefer something else (e.g. a server-side session/token store)?
- Any preferences on config naming and defaults?

Happy to send a PR for whichever direction you prefer.

remidebette:
Closes #242.

## Summary

Once a user completed the OIDC login flow, their session lasted up to ~14 days regardless of any IdP-side change (deactivation, group/MFA changes, password resets, etc.) because only `username` was persisted and Starlette's `SessionMiddleware` fell back to its default cookie TTL. IdP-driven deprovisioning could effectively be ignored for two weeks.

Per @kharkevich's guidance on the issue, this PR implements **option (2) as the default** and **option (3) as opt-in**, then ships the SPA-side plumbing needed to actually make the feature usable.

### Default — token-expiry-based session invalidation (option 2)

- Persist the IdP-issued `expires_at` in the session at OIDC callback time.
- Reject the session in `AuthMiddleware._authenticate_session` once `now >= expires_at - leeway`, forcing the user back through the OIDC flow.
- New config: `OIDC_SESSION_EXPIRY_LEEWAY_SECONDS` (default `30`).
- Sessions that predate this change (no `expires_at`) keep working until their cookie expires — no forced logout at deploy time.

### Opt-in — refresh-token flow (option 3)

- New config: `OIDC_USE_REFRESH_TOKEN` (default `false`).
- When enabled, `offline_access` is appended to the OIDC scope, the refresh token is persisted in the session, and expired sessions are silently refreshed against the IdP via `oauth.oidc.fetch_access_token(grant_type=\"refresh_token\", ...)`. On refresh failure the session is cleared and the user re-logs in.
- Disabled by default because many enterprises require additional approval for `offline_access`, and because refresh tokens then live in the signed-but-not-encrypted session cookie.

### Why three follow-on changes were necessary

Without these, enabling option (2) breaks the SPA mid-session:

1. **Subresource vs document discrimination.** `AuthMiddleware.dispatch` previously redirected (302) any unauthenticated non-`/api` request. After expiry, fetch/XHR and JS chunk loads silently followed the 302 → received HTML → `JSON.parse` and the chunk loader threw. Now `Sec-Fetch-Dest` (with `Accept: text/html` fallback) is used: only document navigations get a 302; subresources get **401** so the SPA can react.
2. **`/static-files` whitelisted.** MLflow's React bundle is served from `/static-files/<path>` with content-addressed hashed filenames and ships publicly on PyPI. Letting it load anonymously avoids `ChunkLoadError` when the session expires mid-use; the next navigation still goes through the redirect.
3. **SPA-side 401 handlers.**
   - **MLflow UI** (we don't own the bundle): a small fetch + `XMLHttpRequest.prototype.send` interceptor is injected via the existing `hack` mechanism. On any 401 it triggers a single `window.location.reload()` → AuthMiddleware redirects to the IdP. Gated by `EXTEND_MLFLOW_REAUTH` (default `true`).
   - **Plugin admin UI**: `core/services/http.ts` now navigates to `/login` on 401 (single-fire, suppressed on `/oidc/ui/auth` to prevent loops). A reload alone wouldn't help here because `/oidc/ui` is in the unprotected-prefix list.

## New configuration

| Variable | Default | Purpose |
|---|---|---|
| `OIDC_SESSION_EXPIRY_LEEWAY_SECONDS` | `30` | Clock-skew leeway applied to the IdP-issued token expiry |
| `OIDC_USE_REFRESH_TOKEN` | `false` | Request `offline_access` and silently refresh expired sessions |
| `EXTEND_MLFLOW_REAUTH` | `true` | Inject the 401-reload helper into MLflow's UI |

## Test plan

- [x] `pytest mlflow_oidc_auth/tests` — 2764 passing
- [x] `yarn test --run` (frontend) — 833 passing
- [x] Manual: silent refresh against JumpCloud (10-min access token, 60-min ID token, refresh token enabled) — verified \"Session for ... refreshed against IdP\" log + uninterrupted requests
- [x] Manual: expiry without refresh token → 401 → SPA-side handler redirects to /login → re-auth completes
- [x] Manual: MLflow UI menu still injects (small unrelated forward-compat fix in commit 1 to match newer MLflow's `#/settings/general?...` sub-routes)

## Notes

- Two pre-existing test failures unrelated to this PR are deselected from the local run (`test_callback_success` redirect-URL assertion, `test_list_deleted_runs_invalid_older_than` schema-mismatch in the local mlflow venv).
- Commit 1 (`fix(ui): match newer MLflow settings sub-routes when injecting menu`) is small and unrelated to #242 but was needed during local testing of this PR. Happy to split it into a separate PR if preferred.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -16,6 +16,8 @@ The application is configured through environment variables, `.env` files, or pl
 | `OIDC_AUDIENCE` | String | None | Expected JWT `aud` claim value (e.g., your client ID or API identifier). When set, bearer tokens are rejected if the `aud` claim doesn't match. Recommended for production to prevent token confusion attacks |
 | `OIDC_PROVIDER_DISPLAY_NAME` | String | `Login with OIDC` | Display name shown on the login page button |
 | `OIDC_GROUPS_ATTRIBUTE` | String | `groups` | Attribute name in the ID token that contains the user's group memberships |
+| `OIDC_SESSION_EXPIRY_LEEWAY_SECONDS` | Integer | `30` | Clock-skew leeway applied to the IdP-issued token expiry. Sessions are rejected once `now >= expires_at - leeway`, forcing the user back through the OIDC login flow so IdP-side changes (deactivation, group changes, MFA enrollment) take effect within the token's lifetime instead of waiting for the cookie TTL |
+| `OIDC_USE_REFRESH_TOKEN` | Boolean | `false` | When `true`, request `offline_access` and persist the refresh token in the session so expired sessions are silently refreshed against the IdP without forcing a visible login. Disabled by default because many enterprises require additional approval for `offline_access` and because refresh tokens are persisted in the signed (but not encrypted) session cookie |
 
 ### Group and Access Control
 
@@ -52,6 +54,7 @@ The application is configured through environment variables, `.env` files, or pl
 | Variable | Type | Default | Description |
 |----------|------|---------|-------------|
 | `EXTEND_MLFLOW_MENU` | Boolean | `true` | Inject sign-in/sign-out links and permission management navigation into MLflow's built-in UI |
+| `EXTEND_MLFLOW_REAUTH` | Boolean | `true` | Inject a small script into MLflow's UI that triggers a full page reload on any 401 response. Without this, an expired session leaves the SPA rendering empty pages until a manual force-reload, because React Router intercepts URL-bar navigations as soft routing |
 | `DEFAULT_LANDING_PAGE_IS_PERMISSIONS` | Boolean | `true` | Use the permissions management page as the default landing page in the admin UI |
 
 ### Feature Flags
--- a/mlflow_oidc_auth/app.py
+++ b/mlflow_oidc_auth/app.py
@@ -193,8 +193,10 @@ def create_app() -> FastAPI:
     for router in get_all_routers():
         oidc_app.include_router(router)
 
-    # Add links to MLFlow UI
-    if config.EXTEND_MLFLOW_MENU:
+    # Inject into MLflow's index.html when the menu links or the re-auth helper
+    # are enabled. Both live in the same hack module to keep injection ordering
+    # predictable.
+    if config.EXTEND_MLFLOW_MENU or config.EXTEND_MLFLOW_REAUTH:
         from mlflow_oidc_auth import hack
 
         app.view_functions["serve"] = hack.index
--- a/mlflow_oidc_auth/config.py
+++ b/mlflow_oidc_auth/config.py
@@ -99,6 +99,15 @@ class AppConfig:
         self.OIDC_GROUPS_ATTRIBUTE = config_manager.get("OIDC_GROUPS_ATTRIBUTE", "groups")
         self.OIDC_AUDIENCE = config_manager.get("OIDC_AUDIENCE")
 
+        # Session re-authentication settings
+        # When True, the session is rejected once the IdP-issued access/ID token expires.
+        # Leeway compensates for clock skew between this server and the IdP.
+        self.OIDC_SESSION_EXPIRY_LEEWAY_SECONDS = config_manager.get_int("OIDC_SESSION_EXPIRY_LEEWAY_SECONDS", default=30)
+        # When True, request `offline_access` and persist the refresh token so the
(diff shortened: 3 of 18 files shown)


remidebette:
@kharkevich this PR should be ready.
Tested locally (Macos, Firefox) with jumpcloud IdP

What do you think?

insunaa:
Can there be an option to set a minimum time before re-authentication is required to hedge against outages of the IdP? Maybe paired with a UI option for admin users "force re-authentication of all users"?

Edit: Alternatively some sort of reachability check for the IdP that keeps expired sessions active while the IdP is not reachable via `http(s)`

Edit2: Does this also kill sessions opened through access keys? I actually can't find where session keys are persisted in the `auth` db and I have use-cases where I expect a session authenticated by access key to stay alive for months and possibly through outages

kharkevich (approved):

kharkevich:
@insunaa could you please open separate discussion, cause I'm going to merge this PR

PR #256, opened 2026-06-12
fix(logout): add clientId parameter for Keycloak

leroyguillaume:
# Summary
This pull request fixes the RP-initiated logout flow, which currently fails against Keycloak (and other strict OpenID providers) with the error page "Missing parameters: id_token_hint".

When redirecting to the provider's end_session_endpoint, the logout handler only appended post_logout_redirect_uri. Since Keycloak 18, whenever post_logout_redirect_uri is present the provider requires either id_token_hint or client_id to also be supplied — otherwise it rejects the logout. As a result, users on Keycloak-backed deployments were left stranded on the provider error page instead of being signed out and redirected back to the MLflow auth page.

The fix sends client_id alongside post_logout_redirect_uri. client_id is used rather than id_token_hint because the session is already cleared at that point and the ID token is not retained for logout; client_id is the spec-sanctioned alternative and is accepted by Keycloak, Auth0, and other providers.

# Key Changes
Logout redirect — mlflow_oidc_auth/routers/auth.py: the end_session_endpoint URL is now built with urlencode({...}) and includes client_id (from config.OIDC_CLIENT_ID) in addition to post_logout_redirect_uri. Both imports (urlencode, config) were already present.
Tests — the two existing logout tests (tests/test_routers_auth.py, tests/routers/test_auth.py) now assert that client_id= is present in the redirect Location header.
Behavior Changes
Keycloak / strict providers: logout now completes successfully and redirects back to the auth page, instead of failing with "Missing parameters: id_token_hint".
Other providers (Auth0, Okta, etc.): no functional change — client_id is an accepted, ignored-or-honored parameter on the end-session endpoint.
Note for operators: the post-logout redirect URI (<mlflow-url>/auth) must be registered in the OIDC client's Valid post logout redirect URIs for the redirect to be honored by Keycloak.

# Testing
81/81 passing in the auth router test suites (tests/test_routers_auth.py, tests/routers/test_auth.py), including the two updated logout tests. black reports no formatting changes.

Fixes the "Missing parameters: id_token_hint" logout failure reported against Keycloak.

### DIFF
--- a/mlflow_oidc_auth/routers/auth.py
+++ b/mlflow_oidc_auth/routers/auth.py
@@ -344,9 +344,20 @@ async def logout(request: Request):
             end_session_endpoint = metadata.get("end_session_endpoint")
 
             if end_session_endpoint:
-                # Redirect to OIDC provider logout with post-logout redirect to auth page
+                # Redirect to OIDC provider logout with post-logout redirect to auth page.
+                # client_id is sent alongside post_logout_redirect_uri because providers
+                # such as Keycloak (>= 18) reject RP-initiated logout with "Missing
+                # parameters: id_token_hint" unless either id_token_hint or client_id is
+                # present. The session is already cleared, so client_id is the reliable
+                # choice here.
                 post_logout_redirect = _build_ui_url(request, "/auth")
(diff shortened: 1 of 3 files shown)


Kilz78:
+1

PR #259, opened 2026-06-24
fix(auth): enforce dynamic experiment URL authorization 

guillaume-thomas:
# Fix: Harden MLflow experiment access control and UI guard

## Motivation
MLflow's SPA-based UI and REST API had several authorization bypass vectors allowing
unauthorized users to access or read data from experiments they don't own:

1. **Direct URL navigation** to experiment pages (hash-based SPA routes were not validated)
2. **Unprotected API endpoints** for model registry operations linked to experiments
3. **Search endpoints** exposing experiment metadata in responses
4. **GraphQL queries** with embedded experiment IDs not subject to authorization

### Backend Authorization Hardening
- **Model Registry Routes** (v2.0):
  - `registered-models/create`: validate experiment_id in prompt context
  - `model-versions/create`: validate against parent model's experiment scope
  - `registered-models/get` & `model-versions/get`: enforce read permission
  - `model-versions/search`: filter results by experiment access

- **Search Endpoints** (v2.0):
  - `runs/search`, `logged-models/search`: validate experiment_ids in payload
  - `registered-models/search`: validate access to experiments in filter tags
  - Extend after_request hooks to strip forbidden experiment references

- **Trace & Dataset APIs** (v3.0):
  - `SearchTraces`: handle experiment_ids in nested location objects
  - Batch GET endpoints: validate against authorized trace IDs
  - Dataset records: enforce experiment linkage permissions

- **GraphQL**:
  - Introspect GraphQL payloads to extract experiment_ids
  - Validate authorization before query execution

### UI Access Guard
- New `experiment_guard.html` snippet injected into MLflow's index.html
- Monitors SPA hash-route changes for experiment navigation
- On unauthorized access:
  - Display styled "Access Denied" overlay (🔒 icon, French message)
  - Hide main content to prevent data leakage
  - Offer "Return to Home" button that reloads the page (full SPA reset)
- Configurable via `EXTEND_MLFLOW_EXPERIMENT_GUARD` (default: enabled)

## Testing
- ✅ All existing tests pass (2800+ tests)
- ✅ 178+ new validator tests added for guarded endpoints
- ✅ Manual verification: unauthorized users now see overlay on restricted experiments
- ✅ Sidebar state is properly reset when redirecting to home

## Configuration
No breaking changes. New behavior is enabled by default but can be disabled via:
```bash
export EXTEND_MLFLOW_EXPERIMENT_GUARD=false
```
## Security Impact

[...]

PR #269, opened 2026-07-24
fix(auth): authorize the traces surface and runs/search (#259)

kharkevich:
Reimplements the genuine authorization gaps from #259 on current `main`. Closes #259 (the original targets the removed `legacy` layout; its client-side `experiment_guard.html` overlay is security theater and is dropped).

## Scope (why most of #259 is gone)

Since #259 was opened, **#264 merged**, making most of it redundant or harmful:
- registered-model/model-version get+create → already proto-bound
- the search before-bindings → **harmful** (`after_request` already filters these gracefully; a before-deny 403s a whole search touching one unreadable item)
- `/graphql` → already enforced by the field-aware `GraphQLAuthorizationMiddleware`
- `experiment_guard.html` overlay → hides SPA DOM *after* the API already answered = security theater

**Three real, still-open gaps** (all default-allow, no `after_request` net) are fixed:

1. **`runs/search`** — bound to the existing `validate_can_read_experiments_from_experiment_ids`.
2. **The entire `/traces/*` read+write surface** — the 6 trace validators were **imported but bound to nothing** (silent no-ops), and two were broken (vacuous-allow on empty scope; path-id routes crashed instead of denying). Rewritten + a new batch validator; bound across ~20 proto classes.
3. Datasets + trace **creation** deferred to focused follow-ups (the latter to avoid gating the hot logging path).

## Two leaks caught by adversarial review (and fixed here)

- **camelCase bypass**: MLflow proto-JSON accepts both snake_case and camelCase, so a location under `mlflowExperiment`/`experimentId` was invisible to the snake-only parser while MLflow still searched it → cross-tenant trace leak. All body fields now read under both spellings.
- **`QueryTraceMetrics` / `CalculateTraceFilterCorrelation`** (`/traces/metrics`, `/traces/calculate-filter-correlation`) take `experiment_ids` and were left unbound → now bound.

## Verification

Enumerated all **52 trace route+method pairs**: every READ route is now bound or a harmless `_not_implemented` (501) stub. Tests cover the vacuous-allow regression, the camelCase bypass, path-id resolution (403 not 500), batch mixed-readability, default-deny on empty, the metrics/correlation bindings, plus a **structural guard that fails if any imported validator is left unbound** (how these traces slipped through originally). Full suite green: **2864 passed**.


guillaume-thomas:
Thank you for this PR folks <3

PR #282, opened 2026-07-25
fix(auth): apply the workspace fallback on artifact-proxy paths (#236)

### LINKED ISSUE #236: [BUG] Artifact proxy returns 403 for workspace users with MANAGE permission
When workspaces are enabled with proxied artifact storage (--serve-artifacts), non-admin users get a 403 Permission denied when uploading artifacts — even if they have MANAGE permission on the workspace. 

  Steps to reproduce:
1. Enable workspaces (MLFLOW_ENABLE_WORKSPACES=true) and proxied artifact storage (--serve-artifacts + --artifacts-destination)
2. Create a workspace (e.g. products) and grant a user MANAGE permission on it
3. As that user, create an experiment in the workspace and start a run 
4. Call mlflow.log_artifact() or mlflow.log_table()   Expected: Artifact uploads go through the proxy using the user's workspace permission.

Actual: 403 Client Error: Forbidden for url: https://mlflow-server/api/2.0/mlflow-artifacts/artifacts/workspaces/products/8/<run_id>/artifacts/file.json Response: {"message":"Permission denied"}

The request correctly goes through the artifact proxy (not direct S3), and the user has MANAGE on the workspace, but the server still rejects it. 
Workaround: Setting DEFAULT_MLFLOW_PERMISSION=EDIT unblocks artifact uploads, but this raises the default permission globally which isn't ideal. 

Possible cause hypothesis: The artifact proxy permission check in validators/experiment.py (_get_experiment_id_from_view_args) appears to expect the experiment ID as the first path segment (<experiment_id>/...), but with workspaces the path starts with workspaces/<name>/<experiment_id>/.... When extraction fails, it falls back to DEFAULT_MLFLOW_PERMISSION instead of checking the user's workspace permission — unlike MLflow's built-in auth which has a workspace-aware fallback

kharkevich:
Fixes #236 — 403 on artifact upload for users whose grant is on the **workspace**. While reproducing it I found the same root cause also permits a **cross-tenant write**, so this is more than a usability fix.

## Root cause

The artifact proxy is the only authenticated path where the workspace is **not** available from the `X-MLFLOW-WORKSPACE` header — MLflow's `http_artifact_repo` never sends it (verified: the module does not reference workspaces at all). The workspace is in the URL instead:

```
/api/2.0/mlflow-artifacts/artifacts/workspaces/products/8/<run_id>/artifacts/file.json
                                     ^^^^^^^^^^^^^^^^^^^^
```

The permission check read the workspace only from the header, so `get_request_workspace()` returned `None`, the workspace fallback was skipped, and resolution fell through to `DEFAULT_MLFLOW_PERMISSION`.

The reporter's hypothesis pointed at `_get_experiment_id_from_view_args`; that part was already fixed by #237 six days after filing. The remaining gap was the **fallback**, which is why the symptom persisted.

## Two failures, opposite directions

Reproduced on `main`, user holding permission only on `own-ws`:

| Request | main | this PR |
|---|---|---|
| upload to own workspace (only a **workspace** grant) | ❌ **403** — the reported bug | ✅ allowed |
| upload to `workspaces/victim-ws/...` (forged path) | ⚠️ **allowed** — cross-tenant write | ✅ **denied** |
| unknown workspace | ⚠️ allowed | ✅ denied |
| plain non-workspace path | allowed | allowed (unchanged) |

The cross-tenant row uses the **shipped default** (`DEFAULT_MLFLOW_PERMISSION=MANAGE`), so it applies out of the box.

## Fix

Parse the workspace out of the artifact path alongside the experiment id — the regex already captured it — and use it for the fallback when the header is absent. Deny when the user holds nothing on that workspace, matching `_apply_workspace_fallback`'s `workspace-deny` rather than falling back to the global default.

The path is caller-controlled, but naming a workspace grants nothing by itself: the check is whether **this** user holds permission on **that** workspace, and MLflow resolves the storage location from the same path, so the check and the write cannot disagree. Only the fallback branch is affected — an explicit experiment-level grant still wins.

## Compatibility
This **tightens** behaviour for deployments on the permissive default: uploads into a workspace path now require a workspace grant. That is the intended isolation property, but it is a behaviour change for anyone relying on `DEFAULT_MLFLOW_PERMISSION=MANAGE`. Flagging explicitly for the release note.

## Tests
11 tests covering both directions plus the unchanged paths (non-workspace, workspaces-disabled, header-present). Mutation-verified: reverting the fallback, making the miss fail open to the default, and dropping the workspaces-disabled guard each fail the suite. Full suite 2952 passing.

Closes #236

### DIFF
--- a/mlflow_oidc_auth/tests/validators/test_artifact_proxy_workspace.py
+++ b/mlflow_oidc_auth/tests/validators/test_artifact_proxy_workspace.py
@@ -1,0 +1,161 @@
+"""Artifact-proxy authorization when workspaces are enabled (issue #236).
+
+The artifact proxy is the only authenticated path where the workspace is NOT
+available from the ``X-MLFLOW-WORKSPACE`` header — MLflow's ``http_artifact_repo``
+does not send it. The workspace is in the URL path instead
+("workspaces/{name}/{experiment_id}/..."), so the permission check has to read it
+from there or the workspace fallback never applies.
+
+Two failures came from that, in opposite directions:
+  * a user whose only grant was on the workspace fell through to
+    DEFAULT_MLFLOW_PERMISSION and got 403 on upload (the reported symptom); and
+  * with the shipped default of MANAGE, any authenticated user could upload into
+    ANY workspace's path, including another tenant's — a cross-tenant write.
+"""
+
+import pytest
+from flask import Flask, request
+
+from mlflow_oidc_auth.config import config
+from mlflow_oidc_auth.validators.experiment import (
+    _parse_artifact_path,
+    validate_can_delete_experiment_artifact_proxy,
+    validate_can_read_experiment_artifact_proxy,
+    validate_can_update_experiment_artifact_proxy,
+)
+
+app = Flask(__name__)
+
+_OWNER = "owner@example.com"
+_OWNED_WS = "own-ws"
+
+
+@pytest.fixture(autouse=True)
+def _clean_permission_cache():
+    """The permission cache is keyed by resource+user and outlives a single test."""
+    from mlflow_oidc_auth.utils.permissions import flush_permission_cache
+
+    flush_permission_cache()
+    yield
+    flush_permission_cache()
+
+
+@pytest.fixture
+def ws_enabled(monkeypatch):
+    monkeypatch.setattr(config, "MLFLOW_ENABLE_WORKSPACES", True)
+    monkeypatch.setattr(config, "PERMISSION_SOURCE_ORDER", ["user", "group"])
+
+
+def _probe(validator, artifact_path, username=_OWNER, header_workspace=None):
+    """Run a proxy validator with the given artifact path and bridge auth context."""
+    from mlflow_oidc_auth.bridge.user import AUTH_CONTEXT_KEY, AuthContext
+
+    with app.test_request_context(path="/proxy", method="PUT"):
+        request.view_args = {"artifact_path": artifact_path}
+        request.environ[AUTH_CONTEXT_KEY] = AuthContext(username=username, is_admin=False, workspace=header_workspace)
+        return validator(username)
+
+
+class TestArtifactPathParsing:
+    def test_workspace_path_yields_experiment_and_workspace(self):
+        with app.test_request_context("/"):
+            request.view_args = {"artifact_path": "workspaces/products/8/run1/artifacts/f.json"}
+            assert _parse_artifact_path() == ("8", "products")
+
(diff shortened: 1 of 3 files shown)


kharkevich:
Closing — **this PR does not fix #236**, and its stated root cause is wrong. Adversarial review caught it; posting the correction rather than merging.

## Where I went wrong

I concluded MLflow does not send `X-MLFLOW-WORKSPACE` on proxied artifact requests by grepping `mlflow/store/artifact/http_artifact_repo.py`, which never mentions workspaces. That search was too narrow: the header is injected one layer up, centrally, in `mlflow/utils/rest_utils.py:165-167`, and `http_artifact_repo.py:91` routes through that same `http_request`.

Verified against a real client:

```
client HAS set_workspace("own-ws")  ->  X-MLFLOW-WORKSPACE = "own-ws"
client has NO workspace             ->  X-MLFLOW-WORKSPACE = None
```

So on a real upload the header IS present, `get_request_workspace()` is truthy, and the new branch — gated on `not get_request_workspace()` — never executes. The PR is a no-op for the flow it targets.

Worse, in the one state that *would* reach it (no workspace context), MLflow itself rejects the request first: `_get_workspace_scoped_repo_path_if_enabled` (`mlflow/server/handlers.py:3499`) raises `INVALID_PARAMETER_VALUE` — *"Active workspace is required for artifact operations"* — and separately raises when the path workspace differs from the request workspace (`:3526`). So the branch could only ever convert a 403 into a 400.

That last check also means the "cross-tenant write" this PR advertised as closing is already backstopped by MLflow: a forged `workspaces/victim-ws/...` path with a mismatched request workspace is rejected upstream. My repro constructed `AuthContext(workspace=None)` by hand — a state MLflow does not accept — which is why the tests passed while proving nothing.

## #236 appears to be already fixed

Re-running the reporter’s scenario the way a real client produces it (header present, MANAGE on the workspace, `DEFAULT_MLFLOW_PERMISSION` without `can_update`):

```
on main (with #237, without this PR):  upload allowed = True
```

The reporter’s own hypothesis was correct — `_get_experiment_id_from_view_args` could not parse `workspaces/{ws}/{id}/...`, so extraction failed and it fell through to `DEFAULT_MLFLOW_PERMISSION`, exactly matching their `DEFAULT_MLFLOW_PERMISSION=EDIT` workaround. That was fixed by #237 on 2026-04-20, six days after the issue was filed, and never confirmed on the issue.

I will follow up on #236 asking the reporter to confirm against current `main` rather than leaving a speculative fix in place.

PR #300, opened 2026-07-30
feat(auth): allow PKCE public clients without a client secret

elifa:
## Summary

`OIDC_CODE_CHALLENGE` (#157) already enables PKCE on the authorization-code flow, but client registration still passed `OIDC_CLIENT_SECRET` unconditionally — so a **public** client, which by definition has no secret, could not be configured at all.

**Root cause:** `_has_required_config()` requires `OIDC_CLIENT_SECRET`, so a secretless config makes `ensure_oidc_client_registered()` return `False`. The app then starts with OIDC quietly unavailable and only a generic startup warning; PKCE is exactly the flow that makes the secret unnecessary.

**Fix:** `oauth.register()` omits `client_secret` when none is set, and `OIDC_CODE_CHALLENGE=S256` satisfies the required-config check in its place.

## Behaviour

| Config | Before | After |
|---|---|---|
| Secret set | registers | unchanged |
| No secret, `OIDC_CODE_CHALLENGE=S256` | login unavailable | registers as a public client |
| No secret, no S256 (id + discovery set) | silent `False` | `ValueError` naming both remedies |
| `OIDC_CODE_CHALLENGE` set to anything but `S256` | passed through | `ValueError` at config load |

Misconfiguration now fails loudly instead of degrading to "no login".

## Tests

- `TestPKCEAndOptionalClientSecret` in `test_oauth.py`: secretless registration omits `client_secret`, secret still passed when present, `code_challenge_method` propagates, `_has_required_config` accepts S256-without-secret and rejects neither-set.
- `test_config.py`: `S256` accepted, non-`S256` rejected, unset disables PKCE.
- `test_auth.py`: login route works for a secretless public client.
- Full suite green locally: **3236 passed**.

## Notes

- Default behaviour is unchanged for every existing confidential-client deployment.
- Docs updated: `OIDC_CLIENT_SECRET` is now "Required unless PKCE", with the public-client setup in `docs/README.md`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -51,6 +51,17 @@ OIDC_ADMIN_GROUP_NAME=mlflow-admin
 SECRET_KEY=your-random-secret-key
 ```
 
+If your OIDC provider issues a public client (no client secret), enable PKCE and omit `OIDC_CLIENT_SECRET`:
+
+```bash
+OIDC_DISCOVERY_URL=https://your-idp.example.com/.well-known/openid-configuration
+OIDC_CLIENT_ID=your-client-id
+OIDC_CODE_CHALLENGE=S256
+OIDC_GROUP_NAME=mlflow
+OIDC_ADMIN_GROUP_NAME=mlflow-admin
+SECRET_KEY=your-random-secret-key
+```
+
 ### 3. Run
 
 ```bash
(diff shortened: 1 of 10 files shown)


elifa:
PKCE has already been merged into main.
This now only contains the support for public clients by changing the validation allowing for missing secret as long as pkce is configured and how those properties are sent to authlib.

PR #301, opened 2026-08-02
feat(auth): support glob patterns in OIDC_GROUP_NAME

### LINKED ISSUE #78: Use patterns for group names allowed to login to mlflow

ValerioSpenn:
## Summary

- allow `OIDC_GROUP_NAME` entries to use case-sensitive shell-style glob patterns
- preserve exact-name behavior and continue synchronizing every group returned by the IdP
- apply the same authorization gate to interactive OIDC login and bearer auto-provisioning
- document the security implications of using `*`

This enables configurations such as `OIDC_GROUP_NAME=mlflow-*` to admit newly created Keycloak groups without updating a ConfigMap or rolling out the server. `OIDC_GROUP_NAME=*` is an explicit opt-in that admits any user with a non-empty group claim. `OIDC_ADMIN_GROUP_NAME` remains exact-match only.

Matching is performed locally against groups already present in the OIDC userinfo/token, so login does not make an additional IdP or Keycloak Admin API request.

Closes #78

## Testing

- `pytest mlflow_oidc_auth/tests/utils/test_groups.py mlflow_oidc_auth/tests/routers/test_auth.py mlflow_oidc_auth/tests/middleware/test_bearer_provisioning.py -q` (75 passed)
- pre-commit checks on changed files (passed)
- full backend run on Windows: 3120 passed; 18 failures in existing Windows path/SQLite migration tests outside this change


ValerioSpenn:
Addressed the Copilot review in db63cc4:

- normalize provider claims and group-detection plugin output before authorization and synchronization
- fail closed for missing, empty, non-string, and mapping values
- ensure `*` only matches non-empty string group names
- cover both interactive login and bearer provisioning with regression tests

Targeted tests: 80 passed. Pre-commit checks: passed.

ValerioSpenn:
Resolved the conflicts with current `main` in merge commit `7225e97`.

Resolution details:
- preserved `OIDC_GROUP_NAME` glob-pattern support while retaining the new configurable username/display-name fields from `main`
- kept group normalization and fail-closed handling for both interactive login and bearer provisioning
- combined the documentation updates, including the empty-configuration warnings

Validation after the merge:
- Black check passed
- 102 targeted group, interactive-auth, bearer-provisioning, and configuration tests passed
- full backend suite: 3,192 passed; the 18 remaining failures are in the existing Windows path/SQLite migration test areas outside this PR

GitHub no longer reports merge conflicts; the branch now contains current `origin/main`.

Issue #305, opened 2026-08-09
auth: establish auth-path performance baseline and query budget


kharkevich:
## Context
Every authenticated request already performs a database round trip. `AuthMiddleware.dispatch` calls `_get_user_admin_status` (`mlflow_oidc_auth/middleware/auth_middleware.py:381` -> `:303`) -> `store.get_user_profile` -> `UserRepository.get_profile` (`mlflow_oidc_auth/repository/user.py:73`), which issues **two queries** (a `load_only` select on `users` plus a `selectinload` for `groups`) on every request, on every path, **uncached** — despite `PERMISSION_CACHE_TTL_SECONDS` and a full pluggable cache backend already existing (`mlflow_oidc_auth/cache/`, `config.py:137,188-190`).

The whole enterprise-identity roadmap is planned on the assumption that this is true — that adding `users.active`, `users.managed_by` and a merged session lookup costs zero additional queries. This task proves or corrects that with numbers.

## Scope
- Measure per-request auth overhead in `AuthMiddleware.dispatch` for session-authenticated, bearer-authenticated and unprotected paths.
- Capture **query count** and wall time on SQLite and PostgreSQL, at 1 / 50 / 500 users and 0 / 20 / 200 groups per user.
- Confirm or correct the specific claim: `_get_user_admin_status` runs on every authenticated request and issues 2 uncached queries via `get_profile`.
- Land a repeatable benchmark (pytest-benchmark, or a script under `scripts/`) runnable with a single command.

## Acceptance criteria
- [ ] Baseline numbers recorded in the repo (query counts + timings per scenario).
- [ ] Benchmark runs from one command and is documented.
- [ ] The 2-query claim is confirmed **or corrected in writing**.
- [ ] A stated budget: no subsequent task may raise per-request auth query count above this baseline.

## Notes
Do this first. Several downstream tasks cite this budget as their acceptance criterion.

### Cross-cutting requirements
- **Back-compat**: a deployment that changes no configuration must see no behavior change.
- **Migrations**: forward and backward, tested on SQLite and PostgreSQL.
- **Audit**: identity, provisioning and revocation decisions emit events via `mlflow_oidc_auth/audit.py`.
- **Performance**: must not increase per-request auth query count above the T0 baseline.
- **Conventions**: Black line-length 160, Google/Sphinx docstrings, `snake_case.py`; frontend `kebab-case.tsx` with co-located tests. Conventional Commits, lowercase subject.

---
[...]

PR #337, opened 2026-08-09
perf(auth): hash token secrets with a cost factor suited to their entropy (#336)

### LINKED ISSUE #336: perf(auth): basic-auth requests spend ~48ms in scrypt password verification on every call
## Context

The auth-path baseline (#305, measured in [#334](https://github.com/mlflow-oidc/mlflow-oidc-auth/pull/334))
turned up one cost that is an order of magnitude larger than everything else on the per-request
auth path, and it is not the database.

A basic-authenticated request costs **~50 ms**, against both SQLite and PostgreSQL, at every point
in the measured matrix ({1, 50, 500} users x {0, 20, 200} groups per user). A session- or
bearer-authenticated request on the same machine costs **1.1–3.4 ms**.

Measured directly, **48.2 ms of that ~50 ms is `werkzeug.security.check_password_hash`** —
`scrypt:32768:8:1`, which is deliberately expensive. The 3 SQL statements the path issues are
noise beside it.

| Scenario | Statements | Median latency |
|---|---:|---:|
| session | 2 | ~1.1–2.8 ms |
| bearer | 2 | ~1.2–3.4 ms |
| **basic** | **3** | **~50 ms** |

Reproduce:

```bash
python scripts/bench_auth_path.py --scenarios session bearer basic
```

## Why it matters

`_authenticate_basic_auth` runs inside `AuthMiddleware.dispatch`, on the request path, before
anything is served. Two consequences:

1. **MLflow CLI and SDK clients pay it per request.** They do not hold a session cookie, so every
   call re-verifies the password from scratch. A client making N tracking calls pays N x 48 ms of
   pure CPU. Service accounts are in the same position.
2. **It is an unauthenticated CPU amplifier.** An attacker who can reach the endpoint can force a
   full scrypt derivation per request with any wrong password — the baseline confirms a wrong
   password still costs the full hash comparison. A modest request rate can saturate CPU. This is
   the standard trade-off for a strong KDF, but right now nothing bounds it.

Note this is **not** a query-count problem, and should not be "fixed" by touching
`UserRepository.get_profile`. The 3-statement count is within the budget recorded in
`docs/performance-baseline.md` and must stay there.

## Scope

Investigate and choose among (not all of these are compatible):

- A short-TTL **verified-credential cache** keyed on username + a hash of the presented secret, so
  a burst of CLI calls verifies once rather than N times. Must be a hash, never the secret; must
  have a short TTL; must be invalidated on password change and on user deletion. The pluggable
  backend in `mlflow_oidc_auth/cache/` already exists.
- **Rate-limiting or a concurrency bound** on the basic-auth path specifically, so failed attempts
  cannot monopolise CPU.
- Revisiting **hash parameters** for service accounts, if their threat model differs from human
  passwords. Weakening the KDF for real user passwords is not on the table.
- Documenting that **bearer tokens are the supported high-volume path**, and steering CLI users
  there, if no mitigation is worth the complexity.

## Acceptance criteria

- [ ] A decision is recorded with a rationale — including "no change, use bearer tokens", if that
      is the outcome.
- [ ] If a cache lands: a test proving a changed password is rejected immediately (not after the
      TTL), and a test proving a wrong password never populates the cache.
- [ ] If a cache lands: no plaintext credential is stored, in memory or in Redis — asserted by a
      test, not by inspection.
- [ ] `pytest mlflow_oidc_auth/tests/perf/test_auth_path_baseline.py` still passes: the basic-auth
      statement count must not rise above 3.
- [ ] `python scripts/bench_auth_path.py --scenarios basic` shows the improvement, with before and
      after numbers in the PR body.

## Notes

Baseline, method and caveats: `docs/performance-baseline.md`. Timings are from one machine with
local databases; the *relative* 25x gap is the durable result, not the absolute milliseconds.

- **Audit**: authentication decisions emit events via `mlflow_oidc_auth/audit.py`.
- **Performance**: must not increase per-request auth query count above the baseline in
  `docs/performance-baseline.md`.
- **Conventions**: Black line-length 160, Google/Sphinx docstrings, `snake_case.py`. Conventional
  Commits, lowercase subject.

---
_Found while establishing the auth-path baseline in #305._

kharkevich:
Closes #336.

## The issue was framed wrong, and the investigation says so

#336 assumed basic auth was slow because of a password-hashing *cost* that had to be worked
around — with a verified-credential cache, rate limiting, or tuning. That framing was mine, and
it was wrong.

**Nothing in this plugin stores a human-chosen password.** Every value ever written to
`users.password_hash` comes from `generate_token()` — 24 characters drawn by `secrets.choice`
from a 62-character alphabet, **142.9 bits of entropy**. There are exactly two writers
(`repository/user.py`, on create and on rotate), and `CreateUserRequest` has no password field:
no endpoint accepts an operator-supplied password.

That makes the 48 ms a category error rather than a tuning problem. scrypt is a memory-hard KDF
whose purpose is making brute-force of *low-entropy* human passwords expensive. Against a
143-bit random token there is nothing to brute-force — at 10^12 guesses/sec, exhausting the space
takes ~10^26 years. The cost bought nothing and was paid on every basic-authenticated request.

So this PR does not add a cache. It corrects the hash method.

## The change

One constant and two call sites:

```python
TOKEN_HASH_METHOD = "pbkdf2:sha256:1000"
```

| method | verify | notes |
|---|---:|---|
| `scrypt:32768:8:1` (old default) | 47.0 ms | |
| `pbkdf2:sha256:600000` | 47.3 ms | no better |
| `pbkdf2:sha256:1000` (**new**) | **0.077 ms** | 610x faster |

Deliberately a constant, not configuration: it is a property of what we store, not a deployment
choice, and the failure mode of setting it wrong is silent.

## Results

Measured with `scripts/bench_auth_path.py` (50 users, 20 groups, 200 iterations, SQLite):

| scenario | before | after | statements |
|---|---:|---:|---:|
| basic | 50.98 ms | **2.81 ms** | 3 → 3 |
| session | 1.31 ms | 1.75 ms | 2 → 2 |
| bearer | 1.56 ms | 2.28 ms | 2 → 2 |
| unprotected | 0.44 ms | 0.55 ms | 0 → 0 |

**Basic auth: 18x faster**, now in the same range as session and bearer. Session/bearer deltas are
run-to-run noise — nothing on those paths changed. The statement budget from #305 is unchanged and
`test_auth_path_baseline.py` still passes.

### A security improvement, not just a speedup

The old cost was itself a **username-enumeration oracle**: a wrong password for an existing user
took ~52 ms (full scrypt) while an unknown user returned in ~0.2 ms (no hash to check). That
250x gap is trivially observable over a network. Measured:

| | wrong password, user exists | unknown user | delta |
|---|---:|---:|---:|
| before | 52.661 ms | 0.191 ms | **52.470 ms** |
| after | 0.284 ms | 0.209 ms | **0.075 ms** |

The oracle shrinks 700x, to well below network jitter. Not tested in CI — a timing assertion
would be flaky — so it is recorded here and in `docs/performance-baseline.md`.

## Back-compat: nothing existing is weakened

Werkzeug encodes the method inside the stored hash and `check_password_hash` dispatches on it, so
hashes written before this change keep verifying **under scrypt, untouched**. No dual-read logic,
no migration script, no schema change.

Existing hashes are deliberately **never re-hashed in place**, even on successful authentication.
A stored secret cannot be distinguished from a hypothetical operator-set password, and silently
re-hashing one down to 1000 iterations would be a real weakening. The cost is that an existing
deployment keeps paying ~50 ms until its tokens are rotated — slower to pay off, but it cannot
weaken anything that exists today. Rotation is already a supported UI action.

Measure that path with `--hash-method scrypt:32768:8:1`, added to the benchmark.

## Tests

`mlflow_oidc_auth/tests/repository/test_user_token_hashing.py`, 14 tests. The ones that matter are
the back-compat and negative cases:

- a pre-change scrypt hash still authenticates, and still **rejects** a wrong secret;
- a successful authentication against a legacy hash leaves the stored hash byte-identical — the
  no-silent-downgrade guarantee, asserted rather than assumed;
- a failed authentication likewise leaves it untouched;
- the new method rejects a wrong secret;
- expiry is still enforced, under both the new and the legacy method;
- hashes are salted (two users, same secret, different hashes), the secret never appears in the
  stored hash, and `get_profile` still redacts.

Also fixed a defect in the #305 benchmark: `_seed` hashed with Werkzeug's default rather than the
repository's method, so it would have kept measuring scrypt forever and reported no improvement.
It now uses `TOKEN_HASH_METHOD`, with `--hash-method` to override.

## Validation

```
pytest mlflow_oidc_auth/tests/ --ignore=.../hooks # 2873 passed
```

No frontend changes. No migration: this changes the format of newly written values in an existing
column, not the schema.

## Acceptance criteria from #336

- [x] **Decision recorded with rationale** — above, and in the `TOKEN_HASH_METHOD` comment. The
      recorded decision is that the premise was wrong: no cache is needed.
- [x] **Changed password rejected immediately, not after a TTL** — no cache exists, so this cannot
      regress; rotation takes effect on the next request by construction.
- [x] **Wrong password never populates a cache** — no cache.
- [x] **No plaintext credential stored anywhere** — asserted by `test_secret_is_not_stored_in_clear`.
- [x] **Basic-auth statement count still 3** — `pytest mlflow_oidc_auth/tests/perf/` passes.
- [x] **Before/after numbers in the PR body** — above.

### DIFF
--- a/docs/performance-baseline.md
+++ b/docs/performance-baseline.md
@@ -86,6 +86,10 @@ Denial paths, asserted in `test_auth_path_baseline.py`:
 Median / p95 milliseconds per request at the ASGI boundary. Both databases are local, so
 these understate a production round trip; they are recorded for orientation, not as a contract.
 
+> The `basic` column is the **T0** measurement, taken before #336 changed how token secrets are
+> hashed. It now costs ~2.8 ms rather than ~50 ms for any secret written after that change; see
+> the findings below. The other three columns are unaffected.
+
 | db | users | groups/user | unprotected | session | bearer | basic |
 |---|---:|---:|---|---|---|---|
 | sqlite | 1 | 0 | 0.41 / 0.48 | 1.09 / 1.28 | 1.28 / 1.74 | 48.91 / 51.23 |
@@ -141,12 +145,23 @@ groups per user adds roughly 1 ms to a session request (~1.3 ms -> ~2.4 ms on SQ
 materialization inside the second query, not extra round trips. Growing from 1 to 500 users
 changes nothing measurable, as the unique index on `users.username` predicts.
 
-**Basic auth is ~25x more expensive than session or bearer, and it is not the database.** A
-basic-auth request costs ~50 ms against both databases, of which a directly measured
-**48.2 ms is `check_password_hash`** — Werkzeug's `scrypt:32768:8:1`, deliberately expensive by
-design. The 3 SQL statements are noise beside it. This is out of scope for the baseline and is
-noted here so it is not mistaken for a database problem: any future work on basic-auth latency
-belongs in password-hash policy or a verified-credential cache, not in query reduction.
+**Basic auth was ~25x more expensive than session or bearer, and it was not the database.** At
+T0 a basic-auth request cost ~50 ms against both databases, of which a directly measured
+**48.2 ms was `check_password_hash`** — Werkzeug's `scrypt:32768:8:1`. The 3 SQL statements were
+noise beside it.
+
+This was fixed in #336, after the baseline was taken. Nothing in this plugin stores a
+human-chosen password: every value in `users.password_hash` comes from `generate_token()`
+(24 characters, 62-character alphabet, ~143 bits of entropy), and no endpoint accepts an
+operator-supplied one. A memory-hard KDF exists to make brute-forcing *low-entropy* passwords
+expensive, so against 143 bits it bought nothing. New hashes use
+`pbkdf2:sha256:1000` (`TOKEN_HASH_METHOD` in `repository/user.py`), taking a basic-auth request
+from **50.98 ms to 2.81 ms median** — into the same range as session and bearer — at an unchanged
+3 statements.
+
+Hashes written before that change keep verifying under their original method and are never
+re-hashed in place, so a deployment that upgrades and rotates nothing still pays the old ~50 ms
+until its tokens are rotated. Measure that path with `--hash-method scrypt:32768:8:1`.
 
 ## Caveats
 
--- a/mlflow_oidc_auth/repository/user.py
+++ b/mlflow_oidc_auth/repository/user.py
@@ -16,6 +16,27 @@ from mlflow_oidc_auth.db.models import SqlGroup, SqlUser
 from mlflow_oidc_auth.entities import User
 from mlflow_oidc_auth.repository.utils import get_user
 
+# Hash method for secrets stored in ``users.password_hash`` (issue #336).
+#
+# Nothing in this plugin stores a human-chosen password. Every value written here comes from
+# ``mlflow_oidc_auth.user.generate_token()`` — 24 characters drawn by ``secrets.choice`` from a
+# 62-character alphabet, about 143 bits of entropy — and no endpoint accepts an operator-supplied
+# password. A memory-hard KDF exists to make brute-forcing *low-entropy* human passwords
+# expensive; against 143 bits there is nothing to brute-force, so the ~48 ms that Werkzeug's
+# default scrypt costs bought no security while being paid on every basic-authenticated request.
+# Verification drops from ~47 ms to ~0.08 ms.
+#
+# Migration is handled by Werkzeug itself: the method is encoded in the stored hash and
+# ``check_password_hash`` dispatches on it, so hashes written before this change keep verifying
+# under scrypt, unchanged. Only newly written hashes use this method — a secret moves over when
+# its token is rotated. Existing hashes are deliberately never re-hashed in place: a stored
+# secret cannot be distinguished from a hypothetical operator-set password, and silently
+# re-hashing one at this cost factor would weaken it.
+#
+# This is intentionally a constant, not configuration. It is a property of what we store, not a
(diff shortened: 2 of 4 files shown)


kharkevich:
Review follow-up.

Added `TestTokenEntropyPremise` (0bb69ce), closing the one review finding attributable to this PR: the new cost factor is justified entirely by `generate_token()`'s entropy, but that function lives in another module and nothing connected the two — shortening the token would have made every stored hash brute-forceable with no test failing. The suite now pins length, alphabet, entropy and non-repetition, and the `TOKEN_HASH_METHOD` comment points at it.

Verified the guard actually fires rather than just passing: shortening the token to 8 characters fails `test_token_length_is_pinned` and `test_token_entropy_clears_the_bar_the_hash_cost_assumes`.

The review also surfaced two pre-existing bugs in `UserRepository.update`, two lines from this diff but not caused by it — a stale expiry surviving token rotation (so a freshly issued token can be born expired), and omitted admin flags silently resetting. Both reproduced against a real store. Filed as #338 rather than folded in here, so this stays a perf change.

PR #339, opened 2026-08-10
fix(auth): make omitted arguments mean "unchanged" in UserRepository.update (#338)

### LINKED ISSUE #338: fix(auth): UserRepository.update mishandles omitted arguments — stale expiry survives token rotation, and admin flags silently reset
## Context

Two defects in `UserRepository.update` (`mlflow_oidc_auth/repository/user.py`), found while
reviewing [#337](https://github.com/mlflow-oidc/mlflow-oidc-auth/pull/337). Both are pre-existing
and neither is caused by that PR — it just edits two lines away. They share one root cause: the
method's handling of "argument not supplied" is inconsistent with its parameter defaults.

Both were reproduced against a real `SqlAlchemyStore` on SQLite, not inferred from reading.

---

## Bug 1 — a rotated token can be born already expired (user-visible)

`update` only assigns the expiration when one is passed:

```python
if password_expiration is not None:
    user.password_expiration = password_expiration
```

`create_access_token` (`routers/users.py:120`) calls
`store.update_user(username=..., password=new_token, password_expiration=expiration)` where
`expiration` is `None` whenever the request body omits it — **the default path**. The previous
expiration therefore survives onto the newly issued token.

`UserRepository.authenticate` checks expiry *before* comparing the hash, so if the old expiration
is in the past, the brand-new token is rejected immediately.

### Reproduction

1. Create an access token with `expiration` set to some date.
2. Wait until it passes (or set it in the past directly).
3. Rotate the token via `PATCH` on the access-token endpoint with **no** `expiration` in the body.
4. Authenticate with the token the endpoint just returned.

Observed on a real store:

```
after rotate: expiration=2026-08-08 23:37:18 (past)
authenticate_user(user, brand_new_token) -> False
```

The user is locked out of basic auth, and rotating again does not help — only supplying an
explicit future expiration does. Whether the admin UI offers that field determines whether an
affected user can self-recover.

### Note on intent

There is a real design question here, which is why this is not a one-line fix: does omitting
`expiration` on rotation mean *"keep the existing policy"* or *"issue a non-expiring token"*?
Today it means neither coherently — it keeps a value that is already spent. My reading is that a
newly issued secret should never inherit a **past** expiry, whichever answer is chosen.

---

## Bug 2 — omitted admin flags silently demote the user (latent)

```python
def update(
    self,
    username: str,
    password: Optional[str] = None,
    password_expiration: Optional[datetime] = None,
    is_admin: Optional[bool] = False,          # <- False, not None
    is_service_account: Optional[bool] = False, # <- False, not None
) -> User:
    ...
    if is_admin is not None:                    # <- False is not None
        user.is_admin = is_admin
    if is_service_account is not None:
        user.is_service_account = is_service_account
```

The guards test for `None`, but the defaults are `False`, so a caller that omits these arguments
**clears** them rather than leaving them alone.

### Reproduction

```
store.user_repo.update(username=<an admin service account>, password=new_token)
-> is_admin=False, is_service_account=False
```

**This is latent, not live.** The only current caller is `SqlAlchemyStore.update_user`
(`sqlalchemy_store.py:356`), which always forwards explicit `None`s and so never trips it. It is a
trap for the next caller: a silent privilege drop with no error and no audit trail. Given
`AGENTS.md`'s deny-by-default posture, a repository method whose default behaviour is "strip the
admin bit" is worth closing before something reaches it.

---

## Scope

- Make "argument omitted" mean one thing consistently. Changing the two defaults to `None` fixes
  Bug 2 and matches `SqlAlchemyStore.update_user`, which already declares them `Optional = None`.
- Decide and document what omitting `expiration` on rotation means, then make Bug 1 follow it. At
  minimum, a newly written secret must not inherit an already-past expiration.
- Check the other `Optional[...] = <non-None>` parameters in the repository layer for the same
  mismatch.

## Acceptance criteria

- [ ] A test proving a token rotated with no `expiration` authenticates successfully when the
      previous expiration is in the past.
- [ ] A test proving `user_repo.update(username=..., password=...)` leaves `is_admin` and
      `is_service_account` unchanged for an admin service account.
- [ ] A negative test proving an explicitly supplied past expiration is still honoured — the fix
      must not turn into "ignore expiry".
- [ ] `pytest mlflow_oidc_auth/tests/repository/ mlflow_oidc_auth/tests/routers/` passes.
- [ ] `pytest mlflow_oidc_auth/tests/perf/` still passes: no new query on the auth path.

## Notes

No schema change is implied — this is assignment logic, not storage.

- **Back-compat**: a deployment that changes no configuration must see no behavior change, beyond
  these two paths starting to behave as documented.
- **Audit**: any change to admin or service-account status should be attributable via
  `mlflow_oidc_auth/audit.py`.
- **Conventions**: Black line-length 160, Google/Sphinx docstrings, `snake_case.py`. Conventional
  Commits, lowercase subject.

---
_Found in review of #337; not introduced by it._

kharkevich:
Closes #338.

Two defects in `UserRepository.update`, both found while reviewing #337 and both sharing one root
cause: what "argument not supplied" meant was inconsistent with the method's own parameter
defaults. Both were reproduced against a real store before being fixed, and the tests were
verified to fail against the unfixed code.

## Bug 1 — a rotated token could be born already expired

`update` only assigned `password_expiration` when one was passed, so the previous token's expiry
survived onto the newly issued one. `authenticate` checks expiry *before* comparing the hash, so
rotating after the old token expired produced a token rejected on its first use.

Reachable through the access-token endpoint whenever the request body omits `expiration` — the
default path for direct API and SDK callers. The React UI always sends one (the submit button is
disabled without it), which is why this never showed up in the browser.

**Fix:** expiry is a property of the *secret*, not of the user. Replacing the secret now replaces
its lifetime with exactly the one supplied, and never inherits:

```python
if password is not None:
    user.password_hash = generate_password_hash(password, method=TOKEN_HASH_METHOD)
    user.password_expiration = password_expiration   # None means "does not expire"
elif password_expiration is not None:
    user.password_expiration = password_expiration
```

Non-rotating updates — group sync, an admin-flag change — still leave the expiry alone.

### The design question from the issue, and how it was settled

#338 flagged an open question: does omitting `expiration` on rotation mean "keep the existing
policy" or "issue a non-expiring token"? Settled as **the latter** — the new token's lifetime is
exactly what was asked for, with no inherited hidden state. Predictability on an auth surface
beats cleverness, and it matches user-creation semantics, where a user provisioned by OIDC login
already gets a token with no expiration.

That choice does widen a credential's lifetime when a script rotates without an expiration, so it
is recorded rather than left silent — the `user.token_rotate` audit event now carries:

```json
{"expiration": null, "expiration_cleared": true}
```

Worth noting for reviewers: the feature's original commit (baaea07) says *"all tokens now have a
lifetime limit"*, which argues for a default lifetime instead. That was never actually true —
every OIDC-provisioned user's initial token has no expiration — so enforcing it now would be a new
behavior change for existing API callers, and belongs in its own issue if wanted.

**Consequence, deliberate:** an expiry can no longer be cleared without also rotating the secret.
Extending the life of an already-issued credential should require issuing a new one.

## Bug 2 — omitted admin flags silently demoted the user

```python
is_admin: Optional[bool] = False,          # default False...
...
if is_admin is not None:                   # ...but the guard tests for None
    user.is_admin = is_admin
```

`store.user_repo.update(username=..., password=...)` therefore set `is_admin=False` and
`is_service_account=False`. Verified: an admin service account came back with both cleared.

**Latent, not live** — the only current caller is `SqlAlchemyStore.update_user`, which always
forwards explicit `None`s. Fixed anyway: it is a silent privilege drop waiting for the next
caller, and the defaults now match both the guards and the store's own signature.

## Tests

`tests/repository/test_user_update_semantics.py` (12) and
`tests/routers/test_users_token_rotate_audit.py` (4).

**Verified the tests bind:** against the unfixed code, 5 of the 12 fail — the three
rotation/expiry cases and both direct-repository flag cases — and all 12 pass after. A test that
would have passed either way is not a regression test.

Negative and boundary cases, since this is an auth path:

- an **explicitly supplied past** expiration is still honoured, so the fix cannot degrade into
  "ignore expiry" — that is the case that would turn a bug fix into a vulnerability;
- flags can still be set `False` **explicitly**, so preserving-on-omission does not make demotion
  impossible;
- a non-rotating update preserves a future expiry;
- `expiration_cleared` is `False` when nothing was dropped, so it is a signal rather than a
  constant.

## Validation

```
pytest mlflow_oidc_auth/tests/ --ignore=.../hooks # 2893 passed
pytest mlflow_oidc_auth/tests/perf/               # 37 passed — #305 query budget intact
```

No schema change: this is assignment logic, not storage. No frontend change — the UI already
always sends an expiration.

### DIFF
--- a/mlflow_oidc_auth/repository/user.py
+++ b/mlflow_oidc_auth/repository/user.py
@@ -180,9 +180,43 @@ class UserRepository:
         username: str,
         password: Optional[str] = None,
         password_expiration: Optional[datetime] = None,
-        is_admin: Optional[bool] = False,
-        is_service_account: Optional[bool] = False,
+        is_admin: Optional[bool] = None,
+        is_service_account: Optional[bool] = None,
     ) -> User:
+        """Update the supplied fields of a user, leaving omitted ones untouched.
+
+        ``None`` means "not supplied" for every parameter but one: the corresponding column is
+        left as it is. The defaults for the two flags previously read ``False`` while the guards
+        below tested for ``None``, so a caller that omitted them silently cleared ``is_admin``
+        and ``is_service_account`` instead of preserving them (issue #338).
+
+        The exception is ``password_expiration``, because expiry is a property of the *secret*
+        rather than of the user:
+
+        * When ``password`` is supplied, the secret is being replaced, so it gets a fresh
+          lifetime — exactly the one passed in, with ``None`` meaning "does not expire". The
+          previous value is never inherited. Inheriting it meant that rotating an already-expired
+          token produced a new token that was rejected on its first use, because ``authenticate``
+          checks expiry before comparing the hash.
+        * When ``password`` is not supplied, the expiry is only changed if one was passed.
+
+        A consequence worth stating: an expiry cannot be cleared without also rotating the
+        secret. That is deliberate — extending the life of a credential that has already been
+        issued should require issuing a new one.
+
+        Parameters:
+            username: Identity key of the user to update.
+            password: New secret. Hashed with :data:`TOKEN_HASH_METHOD`.
+            password_expiration: Expiry for the stored secret. See the semantics above.
+            is_admin: New administrator flag.
+            is_service_account: New service-account flag.
+
+        Returns:
+            User: The updated user entity.
+
+        Raises:
+            MlflowException: If the user does not exist.
+        """
         from werkzeug.security import generate_password_hash
 
         username = normalize_username(username)
@@ -190,7 +224,9 @@ class UserRepository:
             user = get_user(session, username)
             if password is not None:
                 user.password_hash = generate_password_hash(password, method=TOKEN_HASH_METHOD)
-            if password_expiration is not None:
+                # A new secret gets the lifetime it was issued with, never the old one's.
+                user.password_expiration = password_expiration
+            elif password_expiration is not None:
                 user.password_expiration = password_expiration
             if is_admin is not None:
                 user.is_admin = is_admin
--- a/mlflow_oidc_auth/routers/users.py
+++ b/mlflow_oidc_auth/routers/users.py
@@ -3,6 +3,7 @@ from typing import Optional
 
 from fastapi import APIRouter, Body, Depends, HTTPException
 from fastapi.responses import JSONResponse
+from mlflow.exceptions import MlflowException
 
 from mlflow_oidc_auth.audit import emit_audit_event
 from mlflow_oidc_auth.dependencies import check_admin_permission
@@ -97,6 +98,13 @@ async def create_access_token(
 
             try:
                 expiration = datetime.fromisoformat(expiration_str)
+                # An ISO 8601 timestamp carries no offset unless one is written, and both
+                # "2027-01-01" and "2027-01-01T00:00:00" are valid. Comparing a naive datetime
+                # to an aware one raises TypeError, which is not a ValueError and so used to
+                # escape as a 500. This layer deals in UTC — the 'Z' handling above says so —
+                # so read a missing offset as UTC rather than rejecting the request (issue #338).
+                if expiration.tzinfo is None:
+                    expiration = expiration.replace(tzinfo=timezone.utc)
                 now = datetime.now(timezone.utc)
 
                 if expiration < now:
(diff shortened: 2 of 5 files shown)


kharkevich:
Review follow-up: fixed the two 500s in 1e149d2.

**Offset-less ISO expiration returned 500 instead of 400.** `"2027-01-01"` and `"2027-01-01T00:00:00"` are both valid ISO 8601 and parse to naive datetimes; comparing one to the aware `now` raises `TypeError`, which `except ValueError` did not catch. Since this layer already deals in UTC — it rewrites a trailing `Z` — a missing offset is now read as UTC and the request succeeds rather than being rejected. `TypeError` is also caught now, so a bad expiration can never again escape as a 500.

This one mattered beyond the status code, and the interaction is what made it worth fixing here rather than later: an operator hitting the unexplained 500 would naturally retry with `expiration` dropped, and under this PR's new semantics that issues a token which **never expires**. The wrong status code composed with the new default into an accidental permanent credential.

**Unknown target user returned 500 instead of 404.** `get_user_profile` raises rather than returning `None`, so the existing `if user is None` branch was unreachable. The lookup now catches `MlflowException`.

Tests in `tests/routers/test_users_token_rotate_errors.py` (9). Verified they bind: 6 fail against the unfixed router, all 9 pass after. They include the negative cases — a past or beyond-one-year offset-less expiration is still a 400, an explicit offset is still respected rather than trampled by the normalization, genuinely malformed input is still a 400, and no token is issued when the user is unknown. Accepting the format must not stop validating the value.

Full suite: 2902 passed. #305 query budget intact (37 perf tests).

The third finding — the store-level `update_user` not surfacing the new contract — is #340, stacked on this branch since it documents behaviour that only exists once this merges.

PR #88, opened 2025-03-13
feat: support default group permission

grudloffev:
# Summary:

This pull request addresses issue #83. The proposed solution introduces a new environment variable, `DEFAULT_MLFLOW_GROUP_PERMISSION`, that defines the default group permission assigned to assets created by members of a group. This enhancement facilitates scenarios where elevated access is granted within a group (i.e. `READ` or `MANAGE`) while restricting the permissions for users from other groups by using the fallback with lower access (i.e. `NO_PERMISSIONS` or `READ`).

Currently, upon creation of assets, no group permission is assigned, and therefore the fallback defined by the `DEFAULT_MLFLOW_PERMISSION` environment variable is effectively used. To achieve the aforementioned use cases, the group permissions should be assigned manually over the assets, either by the end user or through a separate service.

# Details
- New Feature: Introduced the `DEFAULT_MLFLOW_GROUP_PERMISSION` environment variable.
- Purpose: Default group permission set on experiment and model creation.
- Example configuration:
```bash
export DEFAULT_MLFLOW_PERMISSION="NO_PERMISSIONS"
export DEFAULT_MLFLOW_GROUP_PERMISSION="READ"
```
- Impact: This setting enables use cases where group members have shared access, while ensuring that different groups have limited access.

# Implementation:

- Previous behaviour is maintained when environment variable is unset.
- New environment variable is added to the config (`mlflow-oidc-auth/config.py`).
- New `get_user_groups` util function  (`mlflow-oidc-auth/utils.py`).
- In callback, user groups are added to the session variable (`mlflow-oidc-auth/views/authentication.py`).
- Experiment and Model creation after hooks are modified so that the the default group permission is assigned over the user  groups (`mlflow-oidc-auth/hooks/after_request.py`).
- Model deletion after hook is modified to remove the group permissions (`mlflow-oidc-auth/hooks/after_request.py`).
- The after hook `_set_managed_*_permission` is renamed to `_set_initial_*_permission` to better reflect the new functionality.
- Added `DEFAULT_MLFLOW_PERMISSION` and `DEFAULT_MLFLOW_GROUP_PERMISSION` descriptions in readme and PyPI docs.
- Tests for all covered scenarios of `get_user_groups` (`mlflow-oidc-auth/tests/test_utils.py`).

# Benefits:

- Enhanced Flexibility: Users can now configure default permissions at the group level, providing more granular control over access management.
- Improved Security: By isolating access between groups, organizations can better manage data security and privacy.
[...]

Issue #342, opened 2026-08-10
fix(db): alembic env.py disables every existing logger — plugin and audit logging go silent after the first DB access


kharkevich:
## Summary

`mlflow_oidc_auth/db/migrations/env.py` calls `logging.config.fileConfig(config.config_file_name)`
without `disable_existing_loggers=False`. That parameter defaults to **True**, so every logger
that already exists when migrations run is set to `disabled = True`.

Migrations run on the first access to the lazy store singleton, which happens on the first
authenticated request. By then the plugin's own logger and the audit logger have long since been
created at import time. **Both are silenced for the remaining life of the process.**

`AUDIT_LOG_ENABLED` defaults to `True`, so this affects a default deployment.

## Reproduction

Plugin logger:

```python
from mlflow_oidc_auth.logger import get_logger
log = get_logger()
log.warning("emitted before any migration runs")     # appears
print(log.disabled)                                   # False

from mlflow_oidc_auth.sqlalchemy_store import SqlAlchemyStore
SqlAlchemyStore().init_db("sqlite:////tmp/a.db")

print(log.disabled)                                   # True
log.warning("emitted after the migration ran")        # NOTHING
```

Audit logger — same shape, and this is the part that matters:

```python
emit_audit_event("test.before", actor="probe", ...)   # written
SqlAlchemyStore().init_db(...)
emit_audit_event("test.after",  actor="probe", ...)   # silently dropped
```

Observed output contains `test.before` and does not contain `test.after`.

## Why this matters

`AGENTS.md` requires identity, provisioning and revocation decisions to emit audit events, and
several open issues (#319, #324, and the SCIM work) are specified against that trail existing.

Because migrations run before any auditable action can occur — every one of them requires an
authenticated request, and the first authenticated request is what triggers `init_db()` — the
practical result is that **the audit log records nothing in normal operation**. The events are
constructed and then dropped by a disabled logger, so there is no error and no gap to notice:
the file is simply empty.

This is a broken control rather than a bypass: nothing an attacker does causes it, and there is
no exploit primitive here. It is filed publicly on that basis. It does mean that any incident
review relying on this trail so far has been reading an empty file.

The same mechanism also silences `uvicorn` / `uvicorn.error` / `uvicorn.access`, since the
`[loggers]` section of `alembic.ini` names only `root`, `sqlalchemy` and `alembic`.

## Fix

One line in `mlflow_oidc_auth/db/migrations/env.py`:

```python
[...]

Issue #317, opened 2026-08-09
ui: identity provider picker on the auth page


kharkevich:
## Context
With multiple providers configured, users need to choose one. Today the auth page assumes a single provider.

## Files
- `web-react/src/features/auth/auth-page.tsx` (+ `auth-page.test.tsx`)
- `web-react/src/features/auth/services/`
- `web-react/src/core/hooks/use-auth.ts`
- `web-react/src/core/context/user-provider.tsx`

## Scope
- Fetch the provider list from #316 and render one button per provider.
- **Single provider must behave exactly as today** — auto-redirect when `AUTOMATIC_LOGIN_REDIRECT` is set, no picker chrome.
- Preserve the `?next=` return target through provider selection. The server-side sanitizer at `routers/auth.py:194` already guards this — do not reimplement that logic client-side.
- Distinguish per-provider errors (provider misconfigured vs. user not authorized).
- Show which provider authenticated the current session wherever identity is displayed.
- Build this **provider-type-agnostic** so SAML providers (#330) drop in without rework.

## Acceptance criteria
- [ ] Picker renders for N providers.
- [ ] Single-provider UX byte-for-byte unchanged from today.
- [ ] `next` survives the round trip.
- [ ] Tests cover 0 / 1 / 2+ provider cases.

## Depends on
#316.

### Cross-cutting requirements
- **Back-compat**: a deployment that changes no configuration must see no behavior change.
- **Migrations**: forward and backward, tested on SQLite and PostgreSQL.
- **Audit**: identity, provisioning and revocation decisions emit events via `mlflow_oidc_auth/audit.py`.
- **Performance**: must not increase per-request auth query count above the T0 baseline.
- **Conventions**: Black line-length 160, Google/Sphinx docstrings, `snake_case.py`; frontend `kebab-case.tsx` with co-located tests. Conventional Commits, lowercase subject.

---
_Part of the enterprise identity epic #304. Planned in `auth-expansion-research.md` / `auth-expansion-plan.md`._

kharkevich:
Delivered by #361, whose commits reached `main` through #373 (merge commit 4676a5c); #361 is closed as included.

PR #344, opened 2026-08-10
feat(config): identity provider registry with a legacy back-compat shim (#308)

### LINKED ISSUE #308: config: provider registry with back-compat shim for legacy OIDC_* vars
## Context
Configuration is flat `OIDC_*` singletons (`mlflow_oidc_auth/config.py:99-162`) and cannot express more than one identity provider. Every subsequent task in the enterprise-identity roadmap depends on a registry shape.

## Scope
- Structured provider registry. Each entry: `id`, `type` (`oidc` | `saml` | `k8s`), display name, and per-provider policy:
  - `provisioning`: `jit` | `scim` | `none`
  - `group_sync`: `none` | `first_login` | `every_login`
  - `group_sync_mode`: `additive` | `authoritative`
  - `admin_source`: `claims` | `none` (**never `scim`**)
  - `identity_binding`: `subject` | `email`, plus `allowed_email_domains`
  - `allowed_algorithms` (default `["RS256"]`), `audience` (**required**), `issuer`
- Source: JSON env var and/or file path, resolved through the existing `config_providers` chain so secrets managers keep working.
- **Back-compat is non-negotiable**: with no registry configured, synthesize a single provider `default` from the existing flat `OIDC_*` vars using `jit` / `every_login` / `authoritative` — byte-for-byte today's behavior.
- Startup validation following the existing `_warn_if_*` precedent (`config.py:236-276`). Reject: duplicate ids, `admin_source: scim`, any HMAC algorithm listed alongside a JWKS source, a missing audience, `identity_binding: email` without `allowed_email_domains`, a `saml` provider without the `[saml]` extra installed.

## Acceptance criteria
- [ ] Registry parses from env and file.
- [ ] Legacy flat vars produce an identical `default` provider; existing config tests pass **unmodified**.
- [ ] Every validation case above is covered by a test.
- [ ] No consumer changes in this task — config layer only.

## Notes
`admin_source: scim` is rejected deliberately: Grafana shipped a privilege-escalation flaw where SCIM-reachable group naming controlled admin. Admin group names stay server-side config.

---

### LINKED ISSUE #327: saml: packaging spike — choose library and ship an optional extra
## Context
Both candidate SAML libraries drag a **native** dependency, which is a real problem for a `pip install`-distributed plugin, not a footnote. The answer may change the whole SAML approach, so **start this early** — it is independent of every other task.

## Scope
Compare:
- [`pysaml2`](https://github.com/IdentityPython/pysaml2) — needs the `xmlsec1` **binary** (subprocess). Has XXE CVE history (e.g. CVE-2021-21238); whichever is chosen gets a pin and a monitoring entry.
- [`python3-saml`](https://github.com/SAML-Toolkits/python3-saml) — needs `python-xmlsec` -> `libxmlsec1-dev` **headers at build time**.

Test **actual installability** on every platform this ships to, including the Docker image and a clean manylinux wheel install.

- Ship as an optional extra `[saml]`, following the existing `[aws]` / `[azure]` / `[vault]` pattern in `pyproject.toml`.
- Feature-detect at import; degrade cleanly when absent.
- Document the required OS packages and add them to the image.

## Acceptance criteria
- [ ] One library chosen, with the install evidence written down.
- [ ] `[saml]` installs cleanly on every target platform.
- [ ] The plugin starts normally **without** the extra installed.
- [ ] OS package requirements documented; Docker image updated.

---

kharkevich:
Closes #308.

Flat `OIDC_*` singletons cannot express more than one identity provider, and every later task in
the enterprise-identity epic (#304) needs a registry shape to build against.

**Config layer only — no consumer changes.** Login, callback and token validation are untouched;
the flat variables remain authoritative until something consumes this.

## What it looks like

```json
[
  {
    "id": "okta",
    "type": "oidc",
    "audience": "mlflow",
    "issuer": "https://okta.example.com",
    "provisioning": "jit",
    "group_sync": "every_login",
    "group_sync_mode": "authoritative",
    "admin_source": "claims",
    "identity_binding": "subject",
    "allowed_algorithms": ["RS256"]
  }
]
```

From `AUTH_PROVIDERS` (inline JSON) or `AUTH_PROVIDERS_FILE`, both resolved through the existing
`config_providers` chain so secrets managers keep working. `{"providers": [...]}` is accepted too,
since that is what most people write first.

## Back-compat

With neither set, a single `default` provider is synthesised from the flat variables using
`jit` / `every_login` / `authoritative` — today's behaviour, stated explicitly rather than
implied. **Existing config tests pass unmodified**, as the issue required.

One deliberate asymmetry, and it is the most important decision here: **`audience` is required of
an explicitly configured entry but not of the synthesised one.** `OIDC_AUDIENCE` is optional
today and most installations leave it unset, so holding the shim to the stricter rule would hand
them an empty registry — precisely the break the shim exists to prevent. New configuration is
held to the higher standard; existing configuration is described faithfully, including where it
is weak. Tightening it is a behaviour change and belongs with whatever first consumes the
registry.

I caught this because the first version returned an empty registry on a default install.

## Rejections, and why each one

Invalid entries are **dropped, not repaired** — an entry that survives a check by accident is a
provider someone can authenticate through under a policy nobody wrote.

| Rejected | Why |
|---|---|
| `admin_source: scim` | Whoever controls group naming in the external directory could grant themselves admin. Grafana shipped exactly that. |
| HMAC algorithm + JWKS source | Algorithm confusion: a public verification key replayed as an HMAC secret lets any caller mint an accepted token. |
| Missing `audience` | A token validated with no audience check is valid for every relying party of that issuer. |
| `identity_binding: email` without `allowed_email_domains` | Anyone who can prove any address at any domain could claim a local user. |
| Duplicate `id` | **Every copy**, not just the later one — with two entries claiming an id there is no way to tell which policy was meant, and guessing applies the wrong one. |
| `type: saml` without an implementation | Cannot authenticate against a provider whose protocol is not implemented. |

Dropping rather than raising follows the `_warn_if_*` precedent in `config.py`, for the reason
that file already documents: `AppConfig` is a module-level singleton imported by tooling with
nothing to do with login — Alembic's migration environment, for one — so raising would take that
tooling down over a provider it never reads. It is also the deny-by-default answer: an entry that
cannot be validated does not exist.

A malformed payload or unreadable file falls back to the legacy provider rather than emptying the
registry, so an operator typo cannot lock a working deployment out.

## Tests

`mlflow_oidc_auth/tests/test_provider_registry.py` — 36 tests, one per rule the issue lists, plus
the back-compat and parsing paths. Each rejection test asserts the entry is **gone from the
registry**, not merely flagged.

Also covered: a valid provider survives alongside an invalid one; an entry is never *partially*
accepted (several problems still yield no provider rather than a half-configured one); providers
are immutable, since the registry is shared and a consumer mutating one in place would change
policy for every later request.

## A note on the `[saml]` extra

There is no `[saml]` extra yet — choosing the library is #327's job. Rather than presume that
outcome I check a list of realistic candidate modules, documented so whoever closes #327 has an
obvious place to pin it down. Until then every `type: saml` entry is rejected, which is correct
rather than a gap.

## Validation

```
pytest mlflow_oidc_auth/tests/ --ignore=.../hooks # 2962 passed
pytest mlflow_oidc_auth/tests/test_config.py test_config_providers.py  # 48 passed, unmodified
python -m pyflakes <new files>                    # clean
```

## Merge-order note

Three open PRs also touch `config.py` — #300, #301 and #88. My footprint there is deliberately
small (one import, one assignment, one `_warn_if_*` method) with all the logic in a new module,
so conflicts should be textual at worst.

### DIFF
--- a/mlflow_oidc_auth/config.py
+++ b/mlflow_oidc_auth/config.py
@@ -22,6 +22,7 @@ from dotenv import load_dotenv
 
 from mlflow_oidc_auth.config_providers import config_manager
 from mlflow_oidc_auth.logger import get_logger
+from mlflow_oidc_auth.provider_registry import build_provider_registry
 
 load_dotenv()  # take environment variables from .env.
 logger = get_logger()
@@ -210,11 +211,19 @@ class AppConfig:
         # API documentation settings
         self.ENABLE_API_DOCS = config_manager.get_bool("ENABLE_API_DOCS", default=False)
 
+        # Identity provider registry (issue #308). Built from AUTH_PROVIDERS /
+        # AUTH_PROVIDERS_FILE, falling back to a single "default" provider synthesised from the
+        # flat OIDC_* values above — so a deployment that sets neither behaves exactly as it
+        # does today. Nothing consumes this yet; the flat variables remain authoritative.
+        # Must run after the OIDC_* block, which it reads.
+        self.AUTH_PROVIDERS = build_provider_registry(config_manager, self)
+
         # Run last: these read settings loaded above.
         self._warn_if_resource_creation_restriction_is_inert()
         self._warn_if_default_permission_is_permissive()
         self._warn_if_username_field_unusable()
         self._warn_if_group_name_unusable()
+        self._warn_if_provider_registry_invalid()
 
     @staticmethod
     def _has_usable_entry(field_list) -> bool:
@@ -231,6 +240,23 @@ class AppConfig:
             return any(isinstance(field, str) and field.strip() for field in field_list)
         except TypeError:
             return False
+
+    def _warn_if_provider_registry_invalid(self) -> None:
+        """Report registry entries that were rejected at startup (issue #308).
+
+        Warns rather than raises, for the same reason as the checks around it: ``AppConfig`` is
+        a module-level singleton imported by tooling with nothing to do with login, and taking
+        Alembic down over a malformed provider it never reads would be worse than the
+        misconfiguration. Rejected entries are already absent from the registry, so nothing
+        unvalidated is reachable — this only makes the reason visible.
+
+        Logged at WARNING per entry: an operator who wrote a provider and cannot use it needs
+        the specific reason, and each of these is a security-relevant rejection (a missing
+        audience, an admin source that would allow privilege escalation, an email binding with
+        no domain restriction), not a formatting nit.
+        """
+        for message in self.AUTH_PROVIDERS.errors:
+            logger.warning("Identity provider registry: %s", message)
 
     def _warn_if_username_field_unusable(self) -> None:
         """Warn at startup when OIDC_USERNAME_FIELD or OIDC_DISPLAY_NAME_FIELD is unusable.
--- a/mlflow_oidc_auth/provider_registry.py
+++ b/mlflow_oidc_auth/provider_registry.py
@@ -52,6 +52,31 @@ IDENTITY_BINDINGS = ("subject", "email")
 # replayed as an HMAC secret and any caller can mint a valid token.
 HMAC_ALGORITHMS = ("HS256", "HS384", "HS512")
 
+# Asymmetric algorithms this plugin can actually verify: signatures are checked against keys
+# fetched from the provider's JWKS.
+ASYMMETRIC_ALGORITHMS = (
+    "RS256",
+    "RS384",
+    "RS512",
+    "PS256",
+    "PS384",
+    "PS512",
+    "ES256",
+    "ES384",
+    "ES512",
+    "EdDSA",
+)
+
+# Canonical spelling by uppercase name, so "rs256" and "eddsa" resolve to "RS256" and "EdDSA".
+# Storing the canonical form matters: a consumer comparing against "RS256" would otherwise miss
+# a provider configured as "rs256", and silently fall through to whatever its own default is.
+_CANONICAL_ALGORITHMS = {algorithm.upper(): algorithm for algorithm in ASYMMETRIC_ALGORITHMS + HMAC_ALGORITHMS}
+
+# ``alg: none`` means the token carries no signature at all. Rejected by name and with its own
+# message because it is the single most dangerous value this field can take, and because it is
+# easy to arrive at innocently by copying an example.
+NONE_ALGORITHM = "NONE"
+
 DEFAULT_ALGORITHMS = ("RS256",)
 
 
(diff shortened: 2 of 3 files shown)


kharkevich:
Review follow-up: all three findings fixed in ed39424.

**`allowed_algorithms` was unvalidated, so `alg: none` was accepted.** Confirmed before fixing — `["none"]`, `["NONE"]`, `["totally-made-up"]` and `["RS256", "none"]` all landed in the registry with zero errors, alongside a JWKS source. That is the no-signature bypass: strictly worse than the HMAC-confusion case the original check *did* reject. Algorithms are now checked against the supported set, `none` is rejected by name with its own message, and values are stored canonically so a consumer comparing against `RS256` matches an entry written `rs256` instead of silently falling through to its own default.

Uncomfortable root cause worth naming: my 36 tests covered every rule the issue listed, and the dangerous value was the one the issue didn't think to list. Enumerating the spec is not the same as enumerating the failure modes.

**HMAC is now rejected outright**, not only when the entry itself named `issuer`/`discovery_url`. That gate was evadable by omitting two optional fields while the deployment's flat `OIDC_DISCOVERY_URL` still supplied a key set — the confusion setup restored. This plugin has no symmetric-key verification path, so an HMAC entry cannot work regardless; the only question was whether it failed safely. **This is stricter than #308 specified**, deliberately, and called out in the code so it can be revisited rather than mistaken for the spec.

**Already-parsed registry values are accepted.** `config_providers` are typed `-> Any` and the AWS Secrets Manager provider `json.loads` its whole secret, so a registry stored as nested JSON — the natural way to write one there, and the case #308 explicitly wanted working — arrived as a `list`. The `isinstance(raw, str)` guard treated that as unset and fell back to legacy with **no error at all**: single-provider behaviour, no warning, nothing in `errors`. Lists and dicts are now used directly, and a value of an unusable type is reported instead of ignored.

Also renamed `uses_jwks()` to `has_own_key_source()` and removed it from every security decision, so the two places that disagreed about what "uses JWKS" means no longer both exist.

18 new tests. Verified they bind: **17 fail against the pre-fix module**, all 54 pass after. Full suite 2980 passed; existing config tests still unmodified; #305 budget intact.

kharkevich:
Review follow-up: all three fixed in 56044b7. Two shared a root cause — entry fields consumed without checking they were the primitive the code assumed.

**Unhashable `type` crashed the import.** `INTERACTIVE_BY_DEFAULT.get(provider_type, True)` keyed a dict with an unvalidated JSON value, so `"type": ["oidc"]` raised `TypeError` out of `build_provider_registry`. Since `config = AppConfig()` is module-level, that was not a login-path failure but a total one — verified: `import mlflow_oidc_auth.config` raised, and so did `SqlAlchemyStore().init_db()`, so **migrations died too**. Exactly the outcome this module's docstring argues dropping invalid entries exists to prevent, triggered by one mistyped field, with a traceback naming a dict key instead of the offending config.

Worth naming: this was a regression I introduced in the `interactive` commit. Every earlier check used `in` membership, which tolerates unhashable values; `.get` does not, and I did not notice the difference when adding it.

`_validate_field_types` now runs first and short-circuits, so nothing downstream can `.strip()` a list or key a dict with one. It also covers **`display_name`**, which was unvalidated and would have carried a JSON object into #317's login picker.

**`allowed_algorithms` distinguishes absent from present-but-unusable.** An empty list, a non-list value, or a list with a non-string member are now reported rather than silently becoming RS256 — the same silent-ignore fixed one level up for `AUTH_PROVIDERS` itself. Absent stays quiet, since omitting the field is the documented way to accept the default.

After the fix, with a malformed entry in `AUTH_PROVIDERS`: import succeeds, migrations succeed, and startup logs `provider 'p': 'type' must be a string, got list`.

31 new tests. Verified they bind: **30 fail against the pre-fix module**, all 92 pass after.

One self-inflicted wrinkle worth recording: my first version of the AppConfig test imported `config` while holding a patched `AUTH_PROVIDERS`, which baked that value into the module-level singleton and broke a later test — order-dependent global pollution, the same shape as #342. The test module now forces the singleton into existence at import, from the real environment, before any test can interfere. Verified across repeated randomised runs.

Full suite 3018 passed (x2), hooks per file, existing config tests still unmodified, #305 budget intact.

Issue #333, opened 2026-08-09
db: single Phase 0 schema migration for identity, sessions and lifecycle


kharkevich:
## Context

Splitting Phase 0 into #309, #310 and #311 as originally planned would have each add its own Alembic migration. The revision chain in `mlflow_oidc_auth/db/migrations/versions/` is strictly linear with a single head (`8a9b0c1de234`, `add_workspace_regex_permissions`). Three parallel branches would each set `down_revision = "8a9b0c1de234"`, producing three heads, and `alembic upgrade head` then fails outright.

This issue lands **all Phase 0 schema in one migration** so those three can proceed in parallel on behavior alone.

## Scope

One Alembic revision off `8a9b0c1de234` adding:

**`user_identities`** (for #309)
`id`, `provider_id`, `subject`, `user_id` FK → `users.id`, `created_at`, `last_login_at`. Unique on `(provider_id, subject)`.

**`auth_sessions`** (for #310)
Opaque 256-bit `session_id`, `user_id` FK, `provider_id`, `created_at`, `last_seen_at`, `expires_at`, `revoked_at`, encrypted token material. Indexed on `session_id` and `user_id`.

**`auth_state`** (for #310)
`state`, `nonce`, PKCE verifier, `provider_id`, `relay_state`, `created_at`, `expires_at`. Short-lived; indexed for sweeping.

**Column additions** (for #311)
- `users`: `active` (bool, not null, default true), `managed_by`, `external_id` (nullable, unique when present), `created_at`, `updated_at`
- `user_groups`: `managed_by`
- `groups`: `external_id`, `created_at`, `updated_at`

**Backfill**
- Every existing user gets a `user_identities` row for provider `default`, subject = username. No user is stranded.
- Existing rows: `active=true`, `managed_by=manual` — **`manual`, not `oidc:default`**. Existing memberships were not necessarily claim-derived, and mislabeling them would let the #319 guard lock admins out of their own data.

Add `active` and `managed_by` to the `load_only(...)` list in `mlflow_oidc_auth/repository/user.py:91-98` so they cost no extra query — that row is already fetched on every authenticated request.

## Explicitly out of scope

No behavior. No identity resolution (#309), no session handling (#310), no `active` enforcement or last-admin invariant (#311), no `managed_by` write guard (#319). Schema and backfill only.

## Acceptance criteria

- [ ] `alembic upgrade head` succeeds on SQLite and PostgreSQL from an empty database
[...]

Issue #316, opened 2026-08-09
auth: per-provider login/callback with RFC 9207 mix-up defense


kharkevich:
## Context
There is one `/login` and one `/callback` (`mlflow_oidc_auth/routers/auth.py:238,375`), and per-attempt OAuth state lives in the cookie — so two concurrent login attempts in two tabs clobber each other.

Once this plugin talks to more than one authorization server it becomes exposed to **OAuth mix-up attacks**, which do not exist in a single-provider deployment. RFC 9700 is explicit that a client talking to multiple authorization servers must defend against them: the authorization response never says who sent it, and code + state look identical whichever AS produced them.

## Scope
- Provider-scoped routes `/login/{provider_id}` and `/callback/{provider_id}`. **Preserve legacy `/login` and `/callback`** routed to `default` — redirect URIs already registered at customers' IdPs must not break.
- Per-attempt state moves into #310's `auth_state` table.
- On callback: resolve identity via #309, apply the provider's provisioning policy (#318 if landed, otherwise current behavior), create the #310 session.
- New endpoint listing enabled providers (`id`, display name, type, icon) for #317.
- `auth_status` (`routers/auth.py:448`) reports the provider that authenticated the session rather than the global `OIDC_PROVIDER_DISPLAY_NAME`.

### Mix-up defense — [RFC 9207](https://www.rfc-editor.org/rfc/rfc9207.html)
- Record which issuer started the transaction, in the `auth_state` row.
- Require the returned `iss` parameter to match that issuer **exactly**.
- Fetch tokens from **that issuer's** configured token endpoint — never one derived from the response.
- **Reject a response missing `iss`** when the provider's discovery metadata advertises `authorization_response_iss_parameter_supported`. Otherwise an attacker strips the parameter and recreates exactly the ambiguity RFC 9207 exists to remove.

## Acceptance criteria
- [ ] Two OIDC providers log in independently.
- [ ] Legacy `/login` and `/callback` still work.
- [ ] Concurrent login attempts in two tabs both succeed.
- [ ] Provider list endpoint covered.
- [ ] Mix-up cases from #307 (missing / stripped / mismatched `iss`) all rejected.

## Depends on
#315, #309, #310, #307.

### Cross-cutting requirements
- **Back-compat**: a deployment that changes no configuration must see no behavior change.
- **Migrations**: forward and backward, tested on SQLite and PostgreSQL.
[...]

PR #345, opened 2026-08-10
feat(auth): identity resolution service for multi-provider login (#309)

### LINKED ISSUE #309: auth: add user_identities table and identity resolution service
## Context
`users.username` is the only identity key (`mlflow_oidc_auth/db/models/user.py:19`). With more than one identity provider, two IdPs asserting the same email silently share an account — an account-takeover path. This is also a hard prerequisite for SCIM: `externalId` must key to the same identity as login, or provisioning creates duplicate accounts.

## Scope

> **Updated after #333 and #308 merged.** The schema half of this issue is already done — see the note at the bottom. What remains is the resolution service.

- **Resolution service**: given `(provider_id, subject, claims)` -> an existing user, or a decision to create. Honors the provider's `identity_binding` from the registry (`mlflow_oidc_auth/provider_registry.py`):
  - `subject` — bind on `(provider_id, sub)` only; **never** links across providers.
  - `email` — may link to an existing user by email **only** if the provider is authorized for that email domain via `allowed_email_domains`. Otherwise refuse and log.
- Record `last_login_at` on the resolved identity.
- `username` remains the display and permission key. Nothing is renamed.
- **No consumer changes.** Login and bearer auth keep their current behavior; wiring the service into them is #316's job. This issue lands the service and its tests.

## Acceptance criteria
- [ ] `subject` binding resolves an existing identity and never matches across providers.
- [ ] `email` binding links only within `allowed_email_domains`; a domain outside it is refused and logged.
- [ ] **A cross-provider takeover attempt is refused** — provider B asserting an email already bound to provider A does not reach A's user.
- [ ] An unknown `(provider_id, subject)` yields a create decision rather than a silent match.
- [ ] Identity decisions emit audit events via `mlflow_oidc_auth/audit.py`.
- [ ] `pytest mlflow_oidc_auth/tests/` passes.
- [ ] `pre-commit run --all-files` passes.
- [ ] `pytest mlflow_oidc_auth/tests/perf/` still passes — no new query on the per-request auth path.

## Already done, do not repeat
The table, its constraints and the backfill landed in **#333** (revision `9c0d1e2f3456`):
- `user_identities` `(id, provider_id, subject, user_id, created_at, last_login_at)`, unique on `(provider_id, subject)`, FK to `users.id`, indexed on `user_id`.
- Backfill: every pre-existing user has a row for provider `default`, subject = username.
- ORM model `SqlUserIdentity` in `mlflow_oidc_auth/db/models/identity.py`.

**Do not add another migration for this table.** It exists, so the migration would fail — and it would create a second Alembic head, which is the exact failure #333 was created to prevent.

`identity_binding` and `allowed_email_domains` come from the provider registry landed in **#308**.

## Depends on
#308 (binding policy) and #333 (schema) — both merged.

- **Conventions**: Black line-length 160, Google/Sphinx docstrings, `snake_case.py`. Conventional Commits, lowercase subject.

---

kharkevich:
Closes #309.

`users.username` is the only identity key, so with more than one provider configured two IdPs
asserting the same email would silently share an account. This lands the decision: given
`(provider_id, subject, claims)`, is an existing user reached, should one be created, or is the
attempt refused?

**No consumer changes.** Login and bearer auth are untouched; wiring this in is #316.

## The rules

**`identity_binding: subject`** — a user is reached only through `(provider_id, sub)`. Email is
never consulted, so a token asserting somebody else's address reaches nothing.

**`identity_binding: email`** — may link to an existing user, but only when the address's domain
is in the provider's `allowed_email_domains`, and **never to an account another provider already
owns**. Domain authorisation says a provider may speak for a domain; it does not say it may
inherit an account someone else established.

Two properties hold in both modes:

- **An existing binding wins before any claim is read.** Once `(provider_id, subject)` names a
  user, no claim redirects it, so a second provider cannot talk its way into a bound account.
- **A refusal never carries a username.** A caller that ignores `resolution` and reads `username`
  gets nothing, rather than the account it was just refused. An unauthorised domain is `REFUSED`,
  not "no match" — falling through to create would hand out an account for a domain the operator
  never authorised.

Refusals are audited (`identity.refused`, `status="denied"`): a provider failing to reach an
account it asked for is exactly what an operator investigating a suspected takeover needs. The
reason is recorded; claim contents are not, and a test asserts that.

## Tests

`mlflow_oidc_auth/tests/test_identity_resolution.py` — 29 tests, run against a real store on
SQLite because the exact-match path depends on the unique `(provider_id, subject)` constraint
from #333 and a fake repository would not have it.

The takeover cases are the point:

- a second provider cannot link by email to an account provider A already owns;
- the owning provider *may* still add a second identity for its own user;
- an existing binding beats a conflicting email claim;
- under `subject` binding, a matching email claim reaches nothing;
- the same `sub` under a different provider does not collide (two IdPs both numbering users
  from 1);
- re-linking an identity to a different user is rejected rather than re-pointed, which would be
  a takeover in one call.

## Two defects the tests caught in my own code

- `link()` raised `ValueError` inside the managed session, which wraps every exception into
  `MlflowException(INTERNAL_ERROR)` — so the error code was meaningless and callers could not
  distinguish "unknown user" from "already bound". Now raises `MlflowException` directly with
  `RESOURCE_DOES_NOT_EXIST` / `RESOURCE_ALREADY_EXISTS`.
- `"@example.com"` passed the domain check. It has exactly one `@` and an authorised-looking
  domain, but no local part, so a provider could have claimed an account named for the bare
  domain. The check now requires both halves to be non-empty — it has to be about the whole
  address, not just the half being compared.

## No migration

The table, its constraints and the backfill landed in #333. Adding a revision here would fail
(the table exists) **and** create the second Alembic head that #333 exists to prevent. Verified
after this change: `heads: ['9c0d1e2f3456']`.

I updated #309's scope before starting — it still described the schema work as outstanding, which
would have led exactly there.

## Validation

```
pytest mlflow_oidc_auth/tests/ --ignore=.../hooks # 3047 passed
python -m pyflakes <new files>                    # clean
```

The per-request auth path is unchanged — nothing calls this yet — so the #305 budget is
untouched at 0/2/2/3.

### DIFF
--- a/mlflow_oidc_auth/identity_resolution.py
+++ b/mlflow_oidc_auth/identity_resolution.py
@@ -88,6 +88,15 @@ def _email_from_claims(claims: Dict[str, Any]) -> Optional[str]:
     return email or None
 
 
+def _email_is_verified(claims: Dict[str, Any]) -> bool:
+    """Whether the provider states it has verified the address, strictly.
+
+    Only a literal ``True`` counts. Absent, null, ``"true"`` and every other truthy stand-in are
+    treated as unverified, because this decides whether an address may name an account.
+    """
+    return claims.get("email_verified") is True
+
+
 def _domain_of(email: str) -> Optional[str]:
     """Return the domain part of ``email``, or None when it is not a single usable address.
 
@@ -113,6 +122,7 @@ def resolve_identity(
     claims: Dict[str, Any],
     identity_repo,
     user_lookup,
+    username: Optional[str] = None,
 ) -> IdentityDecision:
     """Decide which local user, if any, an external identity reaches.
 
@@ -124,6 +134,10 @@ def resolve_identity(
         user_lookup: Callable ``(username) -> bool`` answering whether a local user exists.
             Injected rather than taken from the store singleton so this stays a pure decision
             that tests can drive without a database.
+        username: The username the caller derived from the claims using the deployment's
+            configured field list. Optional, but supplying it is what lets email binding notice
+            that this deployment does not name accounts by email — in which case it refuses
+            rather than silently creating a duplicate.
 
     Returns:
         IdentityDecision: What the caller should do. Check ``resolution`` — never read
@@ -140,7 +154,7 @@ def resolve_identity(
         return IdentityDecision(Resolution.MATCHED, username=existing, reason="identity already bound")
 
     if provider.identity_binding == "email":
-        return _resolve_by_email(provider, subject, claims, identity_repo, user_lookup)
+        return _resolve_by_email(provider, subject, claims, identity_repo, user_lookup, username)
 
     # subject binding: an unknown (provider, subject) is a new principal, full stop. It is never
     # matched to an existing user by anything the token says about them.
@@ -153,11 +167,18 @@ def _resolve_by_email(
     claims: Dict[str, Any],
     identity_repo,
     user_lookup,
+    username: Optional[str],
 ) -> IdentityDecision:
     """Link by email, but only within the domains this provider is trusted for."""
     email = _email_from_claims(claims)
     if not email:
         return _refuse(provider, subject, "identity_binding is 'email' but the claims carry no usable email")
+
+    if not _email_is_verified(claims):
+        # An unverified address proves nothing. Refused rather than falling through to CREATE:
+        # creating an account named for an unverified address is pre-registration takeover —
+        # claim the name first, and the real owner logs into the attacker's account later.
+        return _refuse(provider, subject, "the provider did not assert email_verified for this address")
 
     domain = _domain_of(email)
     if not domain:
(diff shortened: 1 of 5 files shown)


kharkevich:
Review follow-up: all three fixed in 31ad720.

**`email_verified` was never consulted.** Confirmed before fixing: with a provider authorised for `example.com`, `{"email": "alice@example.com", "email_verified": False}` returned `LINK` to Alice's account. `allowed_email_domains` bounds *which* domains a provider may speak for and does nothing inside one, so an attacker registering at a legitimately-configured IdP under a colleague's address was handed their account — the standard "Sign in with X" takeover.

Now only a literal `True` counts (`"true"`, `1`, `"yes"` do not — a provider emitting the string `"false"` would otherwise read as verified), and an unverified address is **refused rather than falling through to CREATE**: naming an account after an unproven address is pre-registration takeover, where the attacker claims the name and the real owner later logs into their account.

This is stricter than #309 specified. A provider that cannot assert `email_verified` can no longer use email binding at all and must use `subject` — I judged an opt-out knob to be a footgun on this particular check, but say the word if you want it per-provider.

**The cross-provider guard was advisory.** It lived only in `_resolve_by_email`, and `user_identity_repo` is public on the store — confirmed that `link("rogue", ..., alice)` after `link("okta", ..., alice)` succeeded, giving `['okta', 'rogue']`, exactly what the resolver refuses. So any caller reaching the repository directly bypassed the acceptance criterion, and even a correct caller had a window between resolving and writing for a concurrent login to bind another provider. `link()` now enforces it, with `allow_additional_provider=True` for the deliberate account-linking case so that becomes an explicit decision rather than something that happens by omission.

**Email binding assumed username == email.** Callers may now pass the username they derived from the configured field list; one that is not the address is refused with a message pointing at `identity_binding: subject`. That fixes the silent-duplicate-account case *and* closes a hole my first version had: presenting a verified address of one's own alongside somebody else's username would have reached their account.

16 new tests, and the existing email-binding tests now spell out `email_verified` at every call site so the strictness is visible rather than incidental. Verified they bind: **15 fail against the pre-fix code**, all 45 pass after.

Full suite 3063 passed; hooks per file; #305 budget intact.

PR #347, opened 2026-08-10
feat(auth): register one OAuth client per provider (#315)

### LINKED ISSUE #315: auth: OAuth client registry for multiple providers
## Context
`mlflow_oidc_auth/oauth.py` holds a single module-global authlib client hardcoded as name `"oidc"`, with a one-shot `_oidc_client_registered` flag (`oauth.py:19-20,53-83`). Every consumer reaches it by attribute (`oauth.oidc.*`, e.g. `routers/auth.py:294,305`).

## Scope
- Replace the global single client with a registry of authlib clients keyed by provider id. Keep registration lazy — no network calls at import.
- Route `reset_oauth()` and every `oauth.oidc.*` access through the registry.
- Legacy single-provider config keeps working via #308's `default` provider.

## Acceptance criteria
- [ ] N clients register independently.
- [ ] One provider failing to register does not break the others.
- [ ] Existing oauth tests pass with only the `default` indirection changed.

## Depends on
#308.

---

kharkevich:
Closes #315.

`oauth.py` held a single global authlib client hardcoded as `"oidc"` behind a one-shot
`_oidc_client_registered` flag. That cannot express more than one identity provider — and, more
subtly, cannot express *"provider A registered, provider B failed"*, which is the only state that
matters once there are two.

## What changed

Clients are registered **per provider id**, from the registry landed in #308, with per-provider
state instead of one global flag.

**The legacy name is preserved deliberately.** The `default` provider registers under the authlib
name `"oidc"`, so `oauth.oidc` keeps resolving for every existing call site in `routers/auth.py`
and every test that patches it. Renaming it to `default` would have been a breaking change for no
benefit. A deployment that configures no registry gets the synthesised `default` provider from
the flat `OIDC_*` variables, so nothing about it changes.

```python
ensure_all_clients_registered()   # -> {"okta": True, "broken": False, "entra": True}
```

Each provider is attempted on its own, so one unreachable discovery document does not disable
login for everyone — with several providers, the alternative is one bad entry taking down
authentication for all of them. Startup now logs which registered and which did not, naming them.

## Secrets stay out of the registry

`ProviderConfig` has no `client_secret` field, on purpose: the registry is declarative
configuration that may sit in a JSON file or an environment variable. Additional providers take
their secret from `OIDC_CLIENT_SECRET_<PROVIDER_ID>` resolved through the normal config chain, so
Vault, AWS Secrets Manager and Azure Key Vault keep working. `okta-eu` becomes
`OIDC_CLIENT_SECRET_OKTA_EU` — a provider id that is legal in JSON has to become a legal
environment variable name.

The `default` provider continues to read `OIDC_CLIENT_SECRET`, unchanged.

SAML and `k8s` providers are skipped: neither is driven through an authorization flow.

## Tests

`mlflow_oidc_auth/tests/test_oauth_registry.py` — 17 tests:

- the legacy path: `default` → name `"oidc"`, flat config alone still registers, and registration
  is idempotent (a second call does not re-register);
- **one failing provider does not break the others** — the acceptance criterion, driven by making
  one `register()` raise and asserting the other two still come up;
- a provider with no secret is skipped rather than raising, and an unknown id returns False;
- registration state is genuinely per provider;
- the secret reaches the client from the config chain, and the key derivation handles ids with
  `-` and `.`;
- a guard asserting `ProviderConfig` still has no `client_secret`, so if someone later adds one
  this is where it gets noticed.

**Existing tests: 111 passed untouched.** One changed — it patched the removed
`_oidc_client_registered` global and now patches the per-provider state, which is exactly the
"`default` indirection changed" the issue allows.

## No migration needed

I checked for ORM/migration drift before starting, since several recent PRs added columns:
`compare_metadata` reports 36 differences, **0 of which touch anything added in this epic**, and
the same 36 appear at `d1d44ef` — the commit before any of this work. The drift is pre-existing
(32 `modify_nullable` where models declare NOT NULL but the migrations created the columns
nullable, plus 4 unique-constraint naming mismatches on the workspace regex tables). Worth its own
issue; not touched here.

## Validation

```
pytest mlflow_oidc_auth/tests/ --ignore=.../hooks # 3102 passed (x2 runs)
python -m pyflakes mlflow_oidc_auth/oauth.py      # clean
```

Two pyflakes warnings remain in `app.py` (`typing.Any` unused, `app` redefined). Both are
pre-existing — confirmed identical at `HEAD` — and left alone.

### DIFF
--- a/mlflow_oidc_auth/app.py
+++ b/mlflow_oidc_auth/app.py
@@ -27,7 +27,7 @@ from mlflow_oidc_auth.middleware import (
     WorkspaceContextMiddleware,
     add_fastapi_permission_middleware,
 )
-from mlflow_oidc_auth.oauth import ensure_oidc_client_registered
+from mlflow_oidc_auth.oauth import ensure_all_clients_registered
 from mlflow_oidc_auth.routers import ajax_alias_router, get_all_routers
 
 logger = get_logger()
@@ -57,12 +57,26 @@ async def lifespan(app: FastAPI) -> AsyncIterator[None]:
 
     # Startup: Register OIDC client
     logger.info("Starting MLflow OIDC Auth Plugin...")
-    if ensure_oidc_client_registered():
+    # Every provider is registered independently, so one that is misconfigured or whose
+    # discovery document is unreachable does not disable login for the others (#315).
+    results = ensure_all_clients_registered()
+    registered = [provider_id for provider_id, ok in results.items() if ok]
+    failed = [provider_id for provider_id, ok in results.items() if not ok]
+
+    if registered:
         _oidc_initialized = True
-        logger.info("OIDC client successfully registered at startup")
-    else:
+        logger.info("OIDC client(s) successfully registered at startup: %s", ", ".join(sorted(registered)))
+    if failed:
         logger.warning(
-            "OIDC client registration failed at startup. "
+            "OIDC client registration failed at startup for: %s. "
+            "This may indicate missing configuration (OIDC_CLIENT_ID, OIDC_CLIENT_SECRET, OIDC_DISCOVERY_URL, "
+            "or OIDC_CLIENT_SECRET_<PROVIDER_ID> for an additional provider). "
+            "Those providers will not be available until configuration is corrected.",
+            ", ".join(sorted(failed)),
+        )
+    if not results:
+        logger.warning(
+            "No OIDC client was registered at startup. "
             "This may indicate missing configuration (OIDC_CLIENT_ID, OIDC_CLIENT_SECRET, OIDC_DISCOVERY_URL). "
             "OIDC authentication will not be available until configuration is corrected."
         )
--- a/mlflow_oidc_auth/app.py
+++ b/mlflow_oidc_auth/app.py
@@ -34,6 +34,11 @@ logger = get_logger()
 
 # Global flag to track OIDC initialization status for health checks
 _oidc_initialized: bool = False
+# Per-provider registration outcome from startup. ``_oidc_initialized`` answers the readiness
+# question — can this process serve a login at all — which stays True when one of several
+# providers failed, because failing the startup probe would pull the pod from service while it
+# can still authenticate everyone else. This dict is what says *which* ones are broken (#315).
+_oidc_provider_status: dict[str, bool] = {}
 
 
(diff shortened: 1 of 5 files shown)


kharkevich:
Review follow-up: all three fixed in 2f633e9.

**The fallback overrode an explicit registry.** Confirmed before fixing: a registry holding only a k8s provider, with the old `OIDC_*` variables still in the environment, returned `{'default': True}` and registered `oauth.oidc` against the *old* IdP. That is the normal post-migration state — nobody deletes the old variables the same day — so an operator who deliberately removed browser login would have had it silently restored, with `is_oidc_configured()` reporting true and #317 ready to render a button for a provider absent from their registry.

The fallback is now gated on `source == "legacy"`, which is precisely "no registry configured", and the gate lives in `_client_settings` rather than in `ensure_all_clients_registered`. That matters: `is_oidc_configured()` reaches registration directly, so gating only the bulk path would have left the hole open on the route `routers/auth.py:255` actually uses. There is a test for that specific path.

**Colliding secret keys cross-wired credentials.** `okta-eu` and `okta_eu` are distinct legal ids that both resolve to `OIDC_CLIENT_SECRET_OKTA_EU`, so one provider would have been registered with the other's client secret and failed later as an opaque `invalid_client` from the IdP. Colliding providers are now all refused with an error naming them — with no way to tell which the operator meant, registering either one risks the mix-up. An unrelated provider alongside them still registers.

**Health reported success during a partial failure.** I did *not* make the flag strict: `_oidc_initialized` feeds a Kubernetes **startup probe**, so failing it because one of several IdPs is unreachable would pull the pod from service while it can still authenticate everyone else — the wrong trade for readiness. Instead `oidc_initialized` keeps its "at least one provider works" meaning and the payload now carries a per-provider breakdown, so a broken provider is visible rather than invisible.

10 new tests. Verified they bind: **9 fail against the pre-fix code**, all 27 pass after (x3 runs).

One correction to my own work: `test_flat_config_alone_still_registers` asserted the old fallback using an empty `source="env"` registry. Its premise was wrong — "no registry configured" is `source="legacy"` — so I corrected the test rather than the behaviour, and said so in its docstring.

Full suite 3112 passed (x2); hooks per file; #305 budget intact.

Issue #311, opened 2026-08-09
db: user lifecycle columns (active, managed_by, SCIM meta) with last-admin invariant


kharkevich:
## Context
There is no `active` column, so "deprovision" can only mean hard delete — while Microsoft Entra deprovisions with `PATCH active:false`, not `DELETE`. There is also no record of *which source owns a row*, which is what makes SCIM and claim-based provisioning fight each other.

> **Updated after #333 merged.** The columns, the backfill and the `load_only` addition are already done — see "Already done" below. What remains is **behavior**: enforcement, the invariant, and the recovery path.

## Scope

Three pieces, deliberately kept in one change because they are one safety story: enforcement creates a lockout risk, the invariant prevents it, and the CLI recovers from it. Shipping the first without the other two would be irresponsible.

1. **Enforcement.** `active=false` denies authentication on the session path, the bearer path and basic auth. `active` is already in the `load_only` list, so this must cost **zero** extra queries — the T0 budget (#305) stays at 0/2/2/3.
2. **Last-active-admin invariant**, enforced at the **store layer** so every caller inherits it: refuse any operation that would leave zero active admins.
3. **Break-glass CLI**: a command that restores admin, clears `active=false` and resets `managed_by`, operating directly on the DB and bypassing HTTP auth. `create_app()` already creates a default admin at startup, so the precedent exists. Loudly audited; documented as a runbook step.

Grant handling, ownership transfer and deprovisioning audit belong to #324. The `managed_by` write guard is #319.

## Acceptance criteria
- [ ] Inactive users denied on all three auth paths, with a test each.
- [ ] `pytest mlflow_oidc_auth/tests/perf/` still passes — no added queries on the auth path.
- [ ] Last-active-admin invariant tested on the paths that exist today: delete, deactivate, and demote. (Sync and reconcile are #318/#319 and cannot be tested here.)
- [ ] Break-glass CLI works against a database whose only admin is deactivated.
- [ ] `pytest mlflow_oidc_auth/tests/` passes.
- [ ] `pre-commit run --all-files` passes.

## Already done, do not repeat
Landed in **#333** (revision `9c0d1e2f3456`):
- `users`: `active` (not null, default true), `managed_by`, `external_id` (unique when present), `created_at`, `updated_at`.
- `user_groups`: `managed_by`. `groups`: `external_id`, `created_at`, `updated_at`.
[...]

Issue #310, opened 2026-08-09
auth: server-side session store with real revocation


kharkevich:
## Context
The session lives entirely in the browser cookie: Starlette's `SessionMiddleware` signs the dict with `SECRET_KEY` and the server stores nothing. Consequences:
- **No revocation.** A valid signed cookie stays valid until it expires; there is nothing to delete server-side.
- The refresh token sits in the browser, signed but **not encrypted** (`mlflow_oidc_auth/routers/auth.py:166-192` — its docstring already names this problem).
- No place for per-attempt login state.

This blocks three things: SAML cannot work at all (the IdP's cross-site POST to the ACS endpoint will not carry a `SameSite=Lax` cookie, so state must be correlated server-side via `RelayState`); SCIM `active:false` deprovisioning is cosmetic without something to revoke; and concurrent logins in two browser tabs clobber each other's state.

## Scope
- New tables:
  - `auth_sessions` — opaque 256-bit id, `user_id`, `provider_id`, created/last_seen/expires_at, encrypted token material, `revoked_at`.
  - `auth_state` — short-lived per-attempt state (`state`, `nonce`, PKCE verifier, `provider_id`, `RelayState`), TTL ~10 min, swept.
- Cookie carries **only** the opaque session id. Refresh tokens move into `auth_sessions`, encrypted at rest.
- **Single-query auth lookup**: replace the `_get_user_admin_status` call in `dispatch` with one statement joining `auth_sessions -> users`, returning session validity + `is_admin` (+ `active` once #311 lands). Do not add a second round trip.
- Wrap in the existing cache backend at short TTL, with explicit invalidation on revoke/logout/deactivate. Document the multi-replica caveat: the local backend is per-process, so cross-replica exactness needs Redis.
- Store API: revoke one session; revoke all sessions for a user.
- Decide, document and test the upgrade path for cookies live at deploy time (accept old-format read-only until expiry, **or** force re-login) — do not leave it implicit.

## Acceptance criteria
- [ ] Login creates a DB session; logout and revoke-all kill it immediately.
- [ ] Refresh tokens absent from the cookie; encrypted at rest.
- [ ] T0 benchmark shows per-request auth query count flat or lower.
- [ ] Upgrade path for existing cookies tested.
- [ ] Caching-vs-revocation-latency trade-off documented and configurable.

## Depends on
#305, #306.

### Cross-cutting requirements
[...]

PR #348, opened 2026-08-10
fix(auth): report a deleted user's denial as deletion, and pin it per surface (#306)

### LINKED ISSUE #306: auth: session for a deleted user may still pass authentication
## Context
Suspected live bug, independent of the enterprise-identity roadmap.

`_authenticate_session` (`mlflow_oidc_auth/middleware/auth_middleware.py:215`) authenticates purely from `session["username"]` in the signed cookie — nothing verifies the user still exists. `_get_user_admin_status` (`:303`) catches the "user not found" exception and returns `False`.

Read together, a user deleted mid-session appears to **still pass the authentication middleware**, downgraded to non-admin, with only downstream permission checks blocking them.

**This is not yet confirmed** — it needs an end-to-end trace before anyone asserts it either way.

## Scope
- Delete a user mid-session, then exercise each surface with the still-valid cookie:
  - a FastAPI permission-management route,
  - an MLflow Flask API route (through `AuthAwareWSGIMiddleware` -> `before_request` hooks),
  - `/graphql`.
- Record what each surface actually does today.
- If confirmed: check user existence on the session path (and `active` once #311 lands), and add a regression test per surface.

## Acceptance criteria
- [ ] Behavior documented per surface, with tests, whether or not a gap is found.
- [ ] Any confirmed gap closed with regression coverage.
- [ ] Result reported — it determines whether the server-side session store is a security fix or a feature.

---

kharkevich:
Closes #306.

## The finding

**The gap was real, and it is already closed.**

`_authenticate_session` authenticates purely from `session["username"]` in the signed cookie, and
`_get_user_admin_status` caught the "user not found" exception and returned `False` — so a user
deleted mid-session did pass the authentication middleware, merely downgraded to non-admin, with
only downstream permission checks between them and the application.

The `active` check added by **#311** reads the user's profile on every authenticated request, and
a deleted user cannot be read, so the request is now denied *before* it reaches any surface. I
confirmed the current behaviour by driving the real middleware stack rather than reading the
code: a deleted user's still-valid cookie gets 401 on every surface.

This PR therefore lands the **regression coverage** the issue asked for, plus two reporting fixes
found while verifying.

## Behaviour per surface

Each of the three surfaces the issue named, first shown reachable with a live user so the denial
tests cannot pass for the wrong reason:

| Surface | Live user | Deleted user |
|---|---|---|
| FastAPI route (`/oidc/api/*`) | 200 | **401** |
| Flask route via `AuthAwareWSGIMiddleware` (`/api/2.0/mlflow/*`) | 200 | **401** |
| `/graphql` | 200 | **401** |

All three sit behind `AuthMiddleware`, which `app.py` adds before mounting Flask at `/`, so one
denial covers them. That is a structural claim, and the tests are the evidence — a future
refactor could restore the old behaviour for some of them.

A test also asserts the request **never reaches Flask**: if it did, every route without a
validator would be exposed to a deleted user, and the fix would be one forgotten validator away
from failing.

## Two reporting defects, fixed

Both found by writing the tests, neither visible from reading:

- **A deleted user was audited as `auth.denied_inactive`** and logged as "inactive". Deletion and
  deactivation are different operational events, and conflating them sends an operator
  investigating a lockout to the wrong place. Now `auth.denied_unknown_user`,
  `auth.denied_inactive` and `auth.denied_lookup_error` are distinct, with a test in each
  direction so they cannot drift back together.
- **The missing user was logged at `ERROR`.** A user deleted mid-session is routine, and the
  browser keeps presenting the same cookie until it expires — so this repeated at ERROR on every
  request, about something nobody can act on. It is `INFO` now; ERROR is kept for genuine lookup
  failures.

## The question #306 asked

> Result reported — it determines whether the server-side session store is a security fix or a
> feature.

**A feature.** The authentication gap is closed at the middleware without any server-side session
state, so #310 is not fixing a hole. What it buys is *revocation*: today a signed cookie stays
valid until it expires, and a test here pins the consequence — recreating an account with the
same username restores access to the old session. #310 should change that knowingly rather than
discover it.

## Validation

```
pytest mlflow_oidc_auth/tests/ --ignore=.../hooks # 3124 passed
```

Verified the new tests bind: 2 fail against the pre-fix middleware (the two reporting cases), all
12 pass after. The 10 denial and reachability tests pass either way — correctly, since #311
already closed the gap; they exist so it stays closed.

### DIFF
--- a/mlflow_oidc_auth/middleware/auth_middleware.py
+++ b/mlflow_oidc_auth/middleware/auth_middleware.py
@@ -8,8 +8,10 @@ Authorization (what the user can do) is handled by RBACMiddleware.
 
 from typing import Optional, Tuple
 import base64
+import threading
 import time
 
+from cachetools import TTLCache
 from fastapi import Request, Response
 from fastapi.responses import RedirectResponse, JSONResponse
 from starlette.middleware.base import BaseHTTPMiddleware
@@ -41,6 +43,34 @@ DENIAL_AUDIT_EVENTS = {
     DENIAL_INACTIVE: "auth.denied_inactive",
     DENIAL_LOOKUP_ERROR: "auth.denied_lookup_error",
 }
+
+# A revoked-but-unexpired cookie keeps arriving: a browser tab left open on the SPA issues
+# subresource fetches, polls and telemetry until the cookie expires, and each one is denied. One
+# audit event per request would let a single stale session write thousands of identical entries,
+# burying real events — and would let anyone holding such a cookie dilute the forensic record
+# deliberately. So the *first* denial for a (username, reason) is recorded and repeats within the
+# window are suppressed; the denial itself still happens every time and is still visible in the
+# request log.
+#
+# Bounded as well as time-limited: an attacker rotating usernames must not be able to grow this
+# without limit. Eviction only costs a duplicate audit entry, never a missed denial.
+DENIAL_AUDIT_WINDOW_SECONDS = 60
+_denial_audit_seen: TTLCache = TTLCache(maxsize=1024, ttl=DENIAL_AUDIT_WINDOW_SECONDS)
+_denial_audit_lock = threading.Lock()
+
+
+def _should_audit_denial(username: str, reason: str) -> bool:
+    """Whether this denial is the first of its kind within the window.
+
+    Thread-safe: several ASGI worker threads can be denying the same stale session at once, and
+    ``TTLCache`` is not itself synchronised.
+    """
+    key = (username, reason)
+    with _denial_audit_lock:
+        if key in _denial_audit_seen:
+            return False
+        _denial_audit_seen[key] = True
+        return True
 
 
 def normalize_workspace_header(raw_workspace: Optional[str]) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


kharkevich:
Review follow-up: both fixed in e47df6b.

**One stale cookie was writing one audit event per request.** Measured before fixing: 25 requests carrying a deleted user's cookie produced 25 identical `auth.denied_unknown_user` entries. An open SPA tab issues that many subresource fetches, polls and telemetry calls per page and keeps going until the cookie expires, so a single forgotten tab writes a continuous stream — burying real events in the forensic record, and letting anyone holding a revoked cookie dilute it deliberately with a credential they already had.

Awkwardly, this PR had already fixed exactly that shape for the ERROR log ("logging this at ERROR would fill the log with something nobody can act on") and left it in the audit path, which is the more consequential of the two.

The first denial per `(username, reason)` is now recorded and repeats within 60 seconds suppressed, in a **bounded** TTL cache — an attacker rotating usernames must not grow it without limit, and eviction only ever costs a duplicate entry, never a missed denial — behind a lock, since worker threads can race on the same stale session. Measured after: **25 requests → 1 audit event, and all 25 still 401**. Throttling the record must never throttle the decision, and there is a test for that specifically.

Per-reason and per-user separation is tested too: deactivated-then-deleted for one account produces two events, and two noisy sessions do not hide each other.

**`_get_user_admin_status` reported a deactivated admin as an admin.** No production caller today, but it is named exactly what a future authorization check would reach for, and answered `True` for precisely the deprovisioned admins #311 exists to shut out — the same latent shape as the `UserRepository.update` defaults trap on #339. It now requires the account to be usable as well as admin.

Also dropped a duplicate INFO line; `dispatch` already logs the denial with the path and reason.

7 new tests (19 in the file). Verified they bind: the whole file errors against the pre-fix middleware. Full suite 3131 passed (x2), hooks per file, #305 budget intact.

One note for reviewers: the throttle is deliberately process-wide state, so the tests clear it between cases. Without that, a denial recorded by one test silently suppresses the event another asserts on — which is how the first run failed, and it looked like a missing event rather than leftover state.

PR #352, opened 2026-08-12
feat(auth): enable PKCE by default (#312)

### LINKED ISSUE #312: auth: enable PKCE by default (S256)
## Context
PKCE is supported today but **off unless configured**: `OIDC_CODE_CHALLENGE` defaults to `None` (`mlflow_oidc_auth/config.py:134`) and is passed to authlib as `code_challenge_method` (`mlflow_oidc_auth/oauth.py:76`).

Small, self-contained, and it removes a genuine finding from public comparisons of this plugin against alternatives.

## Scope
- Change the `OIDC_CODE_CHALLENGE` default from `None` to `S256`.
- Verify behavior against a provider that does not support PKCE: the failure must be a clear startup or login error, not a silent 500.
- Keep a documented opt-out.
- Release note — this is a behavior change for anyone whose IdP rejects PKCE.

## Acceptance criteria
- [ ] New deployments use S256 by default.
- [ ] Opt-out works and is documented.
- [ ] Non-PKCE provider produces a clear, actionable error.
- [ ] Upgrade note written.

---

kharkevich:
Closes #312. Last of the Phase 0 foundations in #304.

## The change

PKCE ([RFC 7636](https://www.rfc-editor.org/rfc/rfc7636)) was supported but **off unless `OIDC_CODE_CHALLENGE` was set**, so almost every deployment ran without it. It binds the authorization code to a secret held only by the login attempt that started the flow, so a code intercepted on its way back — a browser history entry, a proxy log, a shared machine, a misconfigured redirect — cannot be redeemed by whoever intercepted it.

`OIDC_CODE_CHALLENGE` now defaults to `S256`.

## ⚠️ Behaviour change

This is the one acceptance criterion that conflicts with the epic's standing "a deployment that changes no configuration sees no behaviour change" rule. The conflict is the point of the issue — an opt-in security default protects nobody — so the rule is broken deliberately, and the cost is bounded: providers that support PKCE are unaffected, and providers that do not now say so clearly instead of failing obscurely.

**Opting out:**

```bash
OIDC_CODE_CHALLENGE=none
```

`none`, `off`, `false`, `no`, `disabled`, `0` and an empty value all disable it — and **log a warning when they do**, so a Helm value that renders blank cannot quietly turn PKCE off while the docs say it is on. `true`, `yes`, `on`, `enabled`, `1` all mean `S256`, so an operator following the upgrade note does not take their server down over the spelling of "on". Anything else is refused at startup.

## `plain` is refused

RFC 7636 defines `plain`, but the pinned authlib emits a challenge for `S256` only — its own docstring says "only S256 is supported", and `client.py` gates `code_challenge` on `self.code_challenge_method == "S256"`. A client configured for `plain` therefore sends an authorization request with **no challenge at all**. Accepting it would report PKCE as enabled while nothing was bound to the code, which is worse than refusing it, so `plain` fails at startup with an explanation.

## How an unsupported provider reports itself

| Situation | What happens |
|---|---|
| Provider advertises `code_challenge_methods_supported` without our method | Login fails **before** the redirect. The log names the provider, the methods it does advertise, and the variable to change |
| Provider advertises nothing ([RFC 8414](https://www.rfc-editor.org/rfc/rfc8414) makes it optional) | Login proceeds — silence is not evidence of absence — and a rejected exchange logs `OIDC_CODE_CHALLENGE=none` as the thing to try, next to the otherwise unexplained `invalid_grant` |
| Discovery unreachable | The check fails open. It is a diagnostic, not an enforcement point; PKCE is applied by authlib either way, and this must never be the thing that breaks a login |

The HTTP response stays generic (`OIDC login is misconfigured; see the server logs`) — `/login` is unauthenticated, and the detailed sentence names an internal registry id.

## Review findings fixed in this PR

A `security-reviewer` pass found four, all fixed above: `plain` silently producing no PKCE at all, an empty variable disabling PKCE with no log line, the disabling-word list rejecting the boolean spellings operators actually use (a self-inflicted outage on the upgrade path), and the 500 body echoing the registry id to an unauthenticated caller.

It also noted that enabling PKCE makes authlib's `DEBUG` log of the per-attempt code verifier live; that is now called out in the docs.

## Validation

```
pre-commit run --all-files                   # passed
pytest -m 'not integration' .../tests        # 3596 passed, 14 skipped (run twice, random order)
pytest mlflow_oidc_auth/tests/perf/          # 37 passed — auth-path budget unchanged
```

28 new tests in `mlflow_oidc_auth/tests/test_pkce_default.py`: the default, every opt-out and affirmative spelling, the `plain` refusal, the typo refusal, the advertised-methods check in both directions, the fail-open paths, that the method actually reaches authlib's `client_kwargs`, and that the two error paths surface to the operator.

No migrations. No frontend changes. Per-request auth query count unchanged — this touches login only.

**One note for the reviewer:** `mlflow_oidc_auth/tests/test_db_utils.py` deletes `mlflow_oidc_auth.config` from `sys.modules`, which leaves the process holding two `config` objects and made these tests fail depending on random test order. The tests here are written to be immune to it, but the underlying trap is still there for the next author. Worth its own issue.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -80,7 +80,7 @@ The plugin uses TTL caches to avoid repeated database lookups on every request. 
 | `OIDC_JWKS_CACHE_TTL_SECONDS` | Integer | `300` | Time-to-live (seconds) for the JWKS key set cache. The OIDC provider's signing keys are fetched once and cached for this duration. This is always a local in-process cache (not affected by `CACHE_BACKEND`) because JWKS data is identical across replicas |
 | `OIDC_HTTP_TIMEOUT_SECONDS` | Integer | `10` | Timeout (seconds) applied to OIDC discovery and JWKS HTTP fetches. Set lower for faster failover when the IdP is unreachable; without a timeout a hung IdP can block request threads until the OS-level TCP timeout (~2 minutes), causing cascading auth failures |
 | `OIDC_VERIFY_SSL` | Boolean | `true` | Verify the OIDC provider's TLS certificate on discovery, JWKS, and token requests. Only set to `false` for providers using self-signed certificates in a trusted network |
-| `OIDC_CODE_CHALLENGE` | String | `S256` | PKCE code-challenge method for the authorization-code flow. `S256` (or `true`/`yes`/`on`/`1`), or `none`/`off`/`false`/`no`/`0` to disable. An unrecognised value is rejected at startup. See [PKCE](#pkce) |
+| `OIDC_CODE_CHALLENGE` | String | `S256` | PKCE code-challenge method for the authorization-code flow. `S256` (or `true`/`yes`/`on`/`1`), or `none`/`off`/`false`/`no`/`0` to disable. An unrecognised value warns and falls back to `S256`. See [PKCE](#pkce) |
 | `PERMISSION_CACHE_TTL_SECONDS` | Integer | `30` | Time-to-live (seconds) for the permission resolution cache. Cached permission decisions expire after this duration. Lower values mean faster propagation of permission changes; higher values reduce database load |
 | `CACHE_BACKEND` | String | `local` | Cache backend for permission and workspace caches. Options: `local` (in-process TTL cache) or `redis` (shared Redis instance). Use `redis` for multi-replica deployments where permission changes must propagate immediately across all replicas |
 | `CACHE_REDIS_URL` | String | None | Redis connection URL. Required when `CACHE_BACKEND=redis`. Example: `redis://localhost:6379/0` or `redis://:password@redis-host:6379/1` |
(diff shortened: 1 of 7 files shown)


kharkevich:
Code review found three issues; all fixed in 78497d3.

**1. The `invalid_grant` hint this PR added could never fire.** authlib removes the per-attempt state from the session — PKCE verifier, nonce, redirect URI — *before* it sends the token request (`starlette_client/apps.py`: `clear_state_data` then `fetch_access_token`). The retry in `_authorize_access_token_with_retry` therefore always failed with `MismatchingStateError`, and **that** error was raised in place of the first attempt's. So a provider rejecting the exchange because of PKCE reported itself to the operator as a CSRF state mismatch — the exact opposite of what the hint was for.

The retry could never have succeeded anyway: the authorization code is single-use, so even with the state intact the provider would refuse the second attempt. It is now one attempt. The JWKS refresh stays and still does its job — a rotated signing key is picked up so the *next* login works — but the error the caller sees is the one that actually happened.

Three tests asserted the impossible second attempt by mocking `authorize_access_token` to succeed on call two. They now assert what the library actually does. The test for the new hint was also stubbing the wrapper itself, which is precisely why the wrapper's error-swallowing was invisible; it now stubs one level below.

**2. An unrecognised `OIDC_CODE_CHALLENGE` no longer stops the process.** `AppConfig` is imported by migration tooling with nothing to do with login, so raising blocked `mlflow-oidc db upgrade` — the upgrade this change asks operators to perform — and `plain` was a legal value in every earlier version, so a live deployment can be carrying one and would have failed to start after upgrading. Unrecognised values (including `plain`) now warn and fall back to `S256`: the secure direction, and the one that keeps the server running. Explicit disabling still works and still warns.

**3.** The parser's docstring no longer claims it can return `"plain"`.

Gate re-run: pre-commit clean, 3597 passed / 14 skipped on two random-order runs, perf 37.

Issue #314, opened 2026-08-09
auth: Kubernetes service-account token provider


kharkevich:
## Context
The automation half of multi-provider support: let a pod authenticate to MLflow with its projected service-account token, while human login stays on Entra/OIDC. These tokens can never be SCIM-provisioned — the corporate IdP does not know they exist — which is why provisioning policy has to be per-provider rather than global.

## Scope
- Provider type `k8s`. Validate projected SA tokens: issuer from the cluster's `/.well-known/openid-configuration`, JWKS at `/openid/v1/jwks`.
- **Audience must be pinned.** An unpinned audience accepts any pod's token for any service — reject such a config outright.
- **Reachability is the likely failure mode, not validation.** Cluster JWKS is frequently not anonymously readable (`system:service-account-issuer-discovery` RBAC) and often unreachable from outside the cluster. Support all three: public discovery, operator-supplied static/inline JWKS, and in-cluster fetch with a custom CA bundle.
- Parse `sub` = `system:serviceaccount:<namespace>:<name>` -> provision as `is_service_account=true`, with a configurable username template and a namespace-derived group.
- SA tokens carry no groups claim, so the existing group gate in `_maybe_provision_bearer_user` (`mlflow_oidc_auth/middleware/auth_middleware.py:160-166`) does not apply. The `k8s` provider needs its own provisioning policy — a namespace allowlist.
- Evaluate the Kubernetes `TokenReview` API as an alternative validator for clusters with no discoverable issuer. Read [opendatahub-io/mlflow-kubernetes-plugins](https://github.com/opendatahub-io/mlflow-kubernetes-plugins) first — same problem, already solved once. Ship it if cheap; otherwise write up why not.

## Acceptance criteria
- [ ] A real projected token from a kind cluster authenticates end to end.
- [ ] Audience confusion between the k8s provider and an OIDC provider is rejected.
- [ ] All three JWKS reachability modes tested.
- [ ] Namespace allowlist enforced.
- [ ] `TokenReview` either implemented or explicitly declined with reasoning.

## Depends on
#313.

### Cross-cutting requirements
- **Back-compat**: a deployment that changes no configuration must see no behavior change.
- **Migrations**: forward and backward, tested on SQLite and PostgreSQL.
- **Audit**: identity, provisioning and revocation decisions emit events via `mlflow_oidc_auth/audit.py`.
- **Performance**: must not increase per-request auth query count above the T0 baseline.
[...]

Issue #307, opened 2026-08-09
auth: adversarial JWT and OIDC callback test suite


kharkevich:
## Context
Multi-issuer token validation and multi-provider login both introduce attack surface that does not exist in a single-provider deployment. This suite is the permanent guard, and it must exist **before** multi-issuer validation ships.

Grounded in [RFC 8725](https://www.rfc-editor.org/rfc/rfc8725.html) (JWT BCP), [RFC 9207](https://www.rfc-editor.org/rfc/rfc9207.html) (AS issuer identification) and RFC 9700 (OAuth 2.0 Security BCP).

## Scope
Test cases, each asserting rejection:
- **Algorithm confusion**: RS256 token re-signed as HS256 using the IdP's public key bytes as the HMAC secret.
- **`alg: none`**.
- **Cross-issuer replay**: a valid token from provider A presented where provider B is expected.
- **`kid` confusion**: `kid` referencing a key belonging to a different configured provider.
- **Audience confusion**: a token minted for the Kubernetes provider's audience accepted at the Entra provider's audience, and vice versa.
- **`jku` / `x5u` header injection**: attacker-supplied key URLs must be ignored, never fetched (SSRF).
- **Callback missing `iss`** where the provider's metadata advertises `authorization_response_iss_parameter_supported`.
- **Callback with stripped `iss`**.
- **Callback with mismatched `iss`** vs. the issuer that started the transaction.

## Acceptance criteria
- [ ] All cases present and passing as rejections.
- [ ] Runs in CI alongside the existing tox envs.
- [ ] Structured so a new provider type inherits the suite rather than re-implementing it.

### Cross-cutting requirements
- **Back-compat**: a deployment that changes no configuration must see no behavior change.
- **Migrations**: forward and backward, tested on SQLite and PostgreSQL.
- **Audit**: identity, provisioning and revocation decisions emit events via `mlflow_oidc_auth/audit.py`.
- **Performance**: must not increase per-request auth query count above the T0 baseline.
- **Conventions**: Black line-length 160, Google/Sphinx docstrings, `snake_case.py`; frontend `kebab-case.tsx` with co-located tests. Conventional Commits, lowercase subject.

---
_Part of the enterprise identity epic #304. Planned in `auth-expansion-research.md` / `auth-expansion-plan.md`._

PR #355, opened 2026-08-12
feat(auth): validate bearer tokens per provider (#313)

### LINKED ISSUE #313: auth: multi-issuer JWKS and per-provider claims validation
## Context
Bearer-token validation is hard-wired to a single provider: the JWKS cache is a single-entry `TTLCache(maxsize=1)` (`mlflow_oidc_auth/auth.py:17`) and `_get_claims_options` (`auth.py:71-85`) reads one `OIDC_ISSUER` / `OIDC_AUDIENCE` pair.

This is the half of multi-provider support that needs no login flow and no UI — it is what makes Kubernetes service-account tokens usable for automation while Entra handles human login.

## Scope
- Replace the single-entry JWKS cache with an issuer-keyed cache. The `force_refresh`-on-`BadSignatureError` rotation retry (`auth.py:100-106`) must become per-issuer or it will thrash a shared cache.
- `_get_claims_options` becomes per-provider.
- Resolve the authenticated token to a user via #309 using the matched provider's binding.

### Security requirements (not optional)
Picking a validator means reading `iss` from an **unverified** token. That is safe only as a lookup key:
1. Exact-match the unverified `iss` against the registry. **Unknown issuer is rejected outright — there is no default fallback validator, ever.**
2. Validate using *only* that provider's JWKS, `allowed_algorithms`, issuer and audience.
3. **Pass `allowed_algorithms` explicitly to `jwt.decode`** (default `["RS256"]`, HMAC family excluded). Authlib does not infer or constrain the algorithm for you — the [Authlib JWT docs](https://docs.authlib.org/en/latest/jose/jwt.html) and [RFC 8725 §3.1](https://www.rfc-editor.org/rfc/rfc8725.html#section-3.1) both require the verifier to state the algorithms it accepts.
4. `aud` is required per provider; an unpinned audience is a config error. Per [RFC 8725](https://www.rfc-editor.org/rfc/rfc8725.html), different uses of JWTs from the same issuer need different `aud` values.
5. `kid` lookups are bound to the resolved provider's JWKS only — never a union across providers. Treat `kid` as an opaque, bounded-charset lookup key; never interpolate it into a query or path.
6. `jku` and `x5u` headers are stripped and ignored — following an attacker-supplied key URL is SSRF.

## Acceptance criteria
- [ ] Two issuers validate independently.
- [ ] A token from issuer A is rejected when presented with issuer B's expected audience.
- [ ] An unknown issuer is rejected with no fallback.
- [ ] JWKS rotation retry is per-issuer.
- [ ] #307 adversarial suite passes.
- [ ] T0 budget holds.

## Depends on
#308, #309, #307 (suite must exist first).

---

kharkevich:
Closes #313. First item of Phase 1 in #304, and the half of multi-provider support that needs no login flow and no UI — it makes Kubernetes service-account tokens usable for automation while Entra handles human login.

## What changed

Validation used a single-entry JWKS cache and one `OIDC_ISSUER`/`OIDC_AUDIENCE` pair. Now the **provider is resolved first**, and everything after comes from that provider alone: its keys, its accepted algorithms, its issuer, its audience.

Selection reads the token's *unverified* `iss`. That is safe for exactly one reason: **the choice can only ever narrow**. An unrecognised issuer selects nothing and the token is refused; a recognised one selects a provider whose constraints are then all enforced against the token that named it. No union of keys is ever offered to the decoder, so a `kid` can only select a key belonging to the issuer the token claims.

**No fallback validator, ever.** A default is precisely what an attacker aims at, since reaching it costs nothing more than an `iss` that matches nothing.

## Back-compat

A deployment with one provider **skips routing entirely** — there is nothing to choose between, and no unverified content is consulted. It keeps using the deployment-wide `_get_oidc_jwks` and the audience/issuer it already configured, which the synthesised `default` provider carries.

One thing had to be corrected to keep that true: the synthesised provider now carries the **whole asymmetric algorithm set**, not the RS256 default of an explicitly written entry. Pre-#313 validation accepted every asymmetric algorithm, so narrowing it would have locked out every ES256 or RS512 deployment on upgrade without them changing a thing. An entry an operator *writes* still defaults to RS256 — there they are stating the policy.

## Key rotation is per issuer

JWKS are cached per provider, keyed on `(id, discovery_url)`, and the `BadSignatureError` retry refreshes only the provider whose signature failed. A shared entry would mean every failure refetched whichever provider happened to be cached, and two issuers on different rotation schedules would evict each other indefinitely. Keying on the source too means repointing a provider at a new IdP does not keep serving the old one's keys for the rest of the TTL.

## Two requirements added to the registry

A `security-reviewer` pass over the diff produced these, and both are now enforced at load with an explanatory error:

- **An explicitly configured token provider must pin `issuer`.** Without it nothing checks `iss`, and a multi-tenant endpoint — Entra `common`, a shared Keycloak realm, a cluster issuer fronting several namespaces — publishes one JWKS for many issuers. A token from any of them carries a valid signature and the right audience; `iss` is the only thing that distinguishes it.
- **It must also name its own `discovery_url`.** Otherwise it inherits the deployment-wide key source and its single-entry cache, and two such providers share one slot.

Also from that pass: the bearer-provisioning gate now asks **the provider that validated the token** whether it pins an audience and issuer, instead of reading flat variables that no longer describe what was enforced — that read could both credit scoping that was never applied and deny provisioning to a registry-only deployment that pins both correctly. And duplicate issuers across entries are rejected rather than resolved by configuration order, which would silently apply the first entry's policy to the second's tokens.

The legacy synthesised provider is exempt from both new requirements: it is built from flat variables that may legitimately be unset, and it is the one provider that must keep working with no configuration change at all.

## Tests

`mlflow_oidc_auth/tests/adversarial/test_multi_issuer.py` — 29 cases that only mean anything once there are two issuers: both validating independently, each refused at the other's audience, an unknown issuer refused with no fallback (including when signed by a configured provider's key), lookalike issuers refused, and rotation proven per-issuer. The #307 `TokenAdversarySuite` runs **twice**, once per provider, against a third unconfigured issuer — the acceptance criterion that a new provider inherits the cases rather than restating them.

## Known limitation, not addressed here

`exp` is still not required. authlib's expiry check is a no-op when the claim is absent, so a token minted without one is accepted indefinitely. That predates this change, and requiring it would break legacy Kubernetes service-account tokens, which have no `exp` — the right place to decide it is #314, where those token semantics are settled. Filing separately.

## Validation

```
pre-commit run --all-files                   # passed
pytest -m 'not integration' .../tests        # 3712 passed, 14 skipped
pytest mlflow_oidc_auth/tests/perf/          # 37 passed — T0 budget unchanged
```

Per-request auth query count is untouched: resolution is a base64 parse and a list scan, no I/O and no query.

### DIFF
--- a/mlflow_oidc_auth/auth.py
+++ b/mlflow_oidc_auth/auth.py
@@ -1,3 +1,5 @@
+import base64
+import json
 import threading
 
 import requests
@@ -8,6 +10,10 @@ from cachetools import TTLCache
 from mlflow_oidc_auth.config import config
 from mlflow_oidc_auth.logger import get_logger
 from mlflow_oidc_auth.provider_registry import ASYMMETRIC_ALGORITHMS
+
+# Provider types whose credentials are bearer tokens verified against a JWKS. SAML asserts
+# identity through a browser POST instead, so it never resolves a token here.
+TOKEN_PROVIDER_TYPES = ("oidc", "k8s")
 from mlflow_oidc_auth.user import create_user, populate_groups, update_user
 
 logger = get_logger()
@@ -25,13 +31,20 @@ logger = get_logger()
 _ACCEPTED_ALGORITHMS = list(ASYMMETRIC_ALGORITHMS)
 _jwt = JsonWebToken(_ACCEPTED_ALGORITHMS)
 
-# JWKS cache: single-entry TTL cache shared across all token validations.
-# TTL is configured via OIDC_JWKS_CACHE_TTL_SECONDS (default 300s).
-# Thread-safe via a lock since multiple ASGI workers may validate concurrently.
+# JWKS cache for the deployment-wide ``OIDC_DISCOVERY_URL``. TTL from
+# OIDC_JWKS_CACHE_TTL_SECONDS (default 300s). Thread-safe via a lock, since multiple ASGI
+# workers validate concurrently.
 _jwks_cache: TTLCache = TTLCache(maxsize=1, ttl=config.OIDC_JWKS_CACHE_TTL_SECONDS)
 _jwks_cache_lock = threading.Lock()
 
 _JWKS_CACHE_KEY = "jwks"
+
+# Per-provider JWKS, keyed by provider id (#313). Separate from the cache above so a rotation
+# retry for one issuer cannot evict another's keys: sharing one entry across issuers would mean
+# every failed signature refetched whichever provider happened to be there, and two issuers with
+# different rotation schedules would thrash each other indefinitely.
+_provider_jwks_cache: TTLCache = TTLCache(maxsize=32, ttl=config.OIDC_JWKS_CACHE_TTL_SECONDS)
+_provider_jwks_lock = threading.Lock()
 
 
 def _get_oidc_jwks(force_refresh: bool = False) -> dict:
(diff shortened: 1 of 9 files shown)


kharkevich:
Correction to the **Back-compat** section above, and the four review findings it came from — fixed in 291911d.

The claim that a single-provider deployment "keeps using the deployment-wide `_get_oidc_jwks`" was **wrong**. The synthesised `default` provider carries `discovery_url` whenever `OIDC_DISCOVERY_URL` is set, so production took the per-provider branch. The test environment sets no discovery URL, so its synthesised provider took the inherited one — meaning every fixture patched `_get_oidc_jwks` and the perf harness primed its cache, and the branch every real deployment runs had no coverage while the covered branch was one production never reached. The two implementations were equivalent, which is precisely why it would have stayed hidden until someone fixed one of them.

There is one implementation now — `_load_jwks`, parameterised by the cache it writes to. Which cache a deployment lands in still depends on whether its provider names a key source; what it *runs* no longer does. Covered by three new tests, including the fact the gap turned on: the synthesised provider carries the configured discovery URL.

Also fixed:

- `TOKEN_PROVIDER_TYPES` was declared in both `auth.py` and `provider_registry.py` — one deciding which providers may validate a token, the other which must pin an issuer and key source. Adding a type to the routing copy alone (the natural edit when wiring #314) would have routed tokens to a provider the registry never required to pin `iss` — the hole this branch added that requirement to close. Now imported.
- `discovery_url` is required of `oidc` providers rather than every token provider. A cluster's keys usually come from the API server's `/openid/v1/jwks` rather than a discovery document, so #314 should state that rule when it knows it; no `k8s` provider can be configured before it lands.
- The module-level `_jwt` decoder is deleted. It was dead once validation became per-provider, while still looking like the live verifier — a change to the accepted algorithms there, which is where the `alg: none` fix landed, would have appeared to take effect and done nothing. `_jwt_for` now narrows within `_ACCEPTED_ALGORITHMS`, so the assertions on it stay meaningful.

Gate re-run: pre-commit clean, 3715 passed / 14 skipped, perf 37.

Issue #318, opened 2026-08-09
auth: per-provider provisioning and group-sync policy


kharkevich:
## Context
Group membership is replaced wholesale from token claims on every login — `update_user` -> `set_user_groups` (`mlflow_oidc_auth/user.py:32-37`), unconditionally. Once SCIM also writes group membership, the next login silently reverts it. This is a design decision, not a bug to patch later.

A **global** SCIM-vs-claims switch is the wrong shape here: Kubernetes SA tokens can never be SCIM-provisioned, so a global switch would force hand-creating every service account.

## Scope
Enforce the #308 policy fields at login and at bearer provisioning:
- `provisioning`: `jit` creates; `scim` and `none` never create — an unknown user is denied with a clear "not provisioned" message (this is the intended enterprise gate, where IdP app-assignment controls access).
- `group_sync`: `none` | `first_login` | `every_login`.
- `group_sync_mode`: `additive` | `authoritative`.
- `admin_source`: `claims` | `none`. **Never SCIM-writable** — `OIDC_ADMIN_GROUP_NAME` stays server-side config.

Touches `mlflow_oidc_auth/user.py:32-37` and `_maybe_provision_bearer_user` (`middleware/auth_middleware.py:113`).

## Acceptance criteria
- [ ] Every policy combination has a test.
- [ ] **Removal is directly tested.** Keycloak's `FORCE` group mapper is add-only in practice ([keycloak#36578](https://github.com/keycloak/keycloak/issues/36578)), so revocation never propagates. `authoritative` must genuinely remove memberships the claims no longer assert.
- [ ] The `default` provider behaves byte-for-byte as today.
- [ ] An unprovisioned user under `scim` gets a clear denial.

## Depends on
#308, #311.

### Cross-cutting requirements
- **Back-compat**: a deployment that changes no configuration must see no behavior change.
- **Migrations**: forward and backward, tested on SQLite and PostgreSQL.
- **Audit**: identity, provisioning and revocation decisions emit events via `mlflow_oidc_auth/audit.py`.
- **Performance**: must not increase per-request auth query count above the T0 baseline.
- **Conventions**: Black line-length 160, Google/Sphinx docstrings, `snake_case.py`; frontend `kebab-case.tsx` with co-located tests. Conventional Commits, lowercase subject.

---
_Part of the enterprise identity epic #304. Planned in `auth-expansion-research.md` / `auth-expansion-plan.md`._

[...]

PR #359, opened 2026-08-12
feat(auth): staged managed_by write guard with dry-run reconcile (#319)

### LINKED ISSUE #319: auth: managed_by write guard with staged enforcement and dry-run reconcile
## Context
Per-provider policy (#318) is a promise nothing verifies. `managed_by` (#311) is the enforcement: a write from one source must not silently clobber a row owned by another.

The failure mode to design against is **admin lockout**, which cannot be recovered from inside the system once every admin is blocked.

## Scope
- Every write path to `users` and `user_groups` checks `managed_by`, refusing to overwrite a row owned by a different source, logging and auditing the refusal (`mlflow_oidc_auth/audit.py`, `emit_audit_event`).
- Admin override exists and is audited — operators must be able to break glass.
- **Three-state enforcement**: `MANAGED_BY_ENFORCEMENT = off | report | enforce`, defaulting to `report` for at least one minor release. In `report`, the guard evaluates, logs and audits what it *would* have rejected, and changes nothing. Ship the telemetry before the enforcement.
- **Dry-run reconciliation CLI**: prints the exact diff; applies nothing without an explicit flag. It **never runs implicitly at startup** — [grafana#73752](https://github.com/grafana/grafana/issues/73752) was precisely a silent runtime branch that reset existing users.
- **Reversibility**: reconciliation records prior `managed_by` values so a mistaken run can be rolled back within a retention window.
- Repair path for when a source is turned off (rows revert to `manual`).

## Acceptance criteria
- [ ] Cross-source writes refused and audited under `enforce`; logged but permitted under `report`.
- [ ] Admin override works and is logged.
- [ ] Switching a provider's policy mutates nothing until reconciliation is explicitly run.
- [ ] Dry-run output matches what apply actually does.
- [ ] Rollback restores prior ownership.
- [ ] Lockout scenarios explicitly tested — never relies on review.

## Depends on
#318, #311.

---

kharkevich:
Closes #319. Per-provider policy (#318) says what each source *should* write; this is what stops one source silently overwriting a row another owns when it writes something else.

## Staged, because the failure mode is lockout

`MANAGED_BY_ENFORCEMENT` has three states and defaults to the middle one:

| Value | Behaviour |
|---|---|
| `off` | No evaluation |
| `report` | **Default.** Audits `user.ownership_conflict`; the write proceeds |
| `enforce` | The write is refused |

A guard that refuses writes cannot be recovered from inside a system that has just refused the write that would fix it — so the telemetry exists a release before the enforcement, and an operator can look at their own traffic before turning it on. **An administrator action is permitted in every mode and always audited**; the admin API writes as an explicit override.

## Ownership never changes implicitly

Not at startup, not on a configuration change, not as a side effect of a login — Grafana shipped a silent runtime branch that reset existing users, and the lesson is that this is an operator action with a diff they read first.

```bash
mlflow-oidc db reconcile-ownership --url "$DB" --from-owner scim --set-owner manual            # dry run
mlflow-oidc db reconcile-ownership --url "$DB" --from-owner scim --set-owner manual --apply --journal /tmp/own.json
mlflow-oidc db restore-ownership   --url "$DB" --journal /tmp/own.json --apply                 # rollback
```

The dry-run diff and the applied diff come from the same query. This is also the repair path when a source is turned off: `--from-owner` it, `--set-owner manual`.

## The review finding that mattered

The guard was wired into `UserRepository.update`, but **every production caller goes through `store.update_user`**, which carried no attribution. So under `enforce` every write to a directory-owned row was refused — *including that directory's own sync* — and `create_user` caught the refusal, fell into its create branch and reported "user already exists". That is a lockout of exactly the users the guard exists to protect, with an error pointing somewhere else.

My tests passed `written_by=` explicitly in every case, so the production shape — the caller that omits it — had no test. Attribution now reaches the store, the login path writes as `oidc:<provider>`, the admin API writes as an override, and the case is tested *through* the caller.

## Also from that review

- `Enforcement` is a `str`-Enum, so `enforcement is Enforcement.ENFORCE` is False for any caller holding the raw configured string — the guard would have been silently off while every log line said `enforce`. Now `==`, and an unrecognised value reports rather than falling through to "allowed".
- The permitted-conflict audit is emitted **after** the commit, so it cannot claim a cross-source write that a rollback undid — the same reason the session-revocation audit moved in #351.
- `--set-owner` is validated against owners a source actually presents. A typo under `enforce` means every writer conflicts with it forever, and it is usually typed by someone already repairing a lockout.
- Re-owning every row requires `--all`; a username is matched as stored rather than as typed; a journal is never overwritten; and a restore leaves rows deliberately re-owned since it was written, naming them.

## Validation

```
pre-commit run --all-files              # passed
pytest -m 'not integration' .../tests   # 3909 passed, 14 skipped
pytest mlflow_oidc_auth/tests/perf/     # 37 passed — no query added to the auth path
```

46 tests in `test_ownership_guard.py`: the three modes, break glass in every mode, the guard on the write path, the lockout scenarios, the reconcile/restore round trip and the CLI footguns.

## Known, not addressed

- **The guard covers `update` only.** `set_user_groups` and `delete` are unguarded, and `user_groups.managed_by` exists but is never read — so a second provider can still replace group membership, which is where permissions actually live. That is the next slice and wants its own issue rather than a quiet extension of this one; `report` mode makes the gap visible in the meantime.
- Bulk ownership changes are audited with `actor="cli"` and no operator identity.
- The bearer provisioning path logs an ownership refusal as "may already exist".

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -130,7 +130,17 @@ so the telemetry exists a release before the enforcement does. Run on `report`, 
 `user.ownership_conflict` events you actually get, then move to `enforce`.
 
 An administrator action is permitted in every mode and always audited. That is deliberate: an
-operator who cannot fix ownership from the admin UI is left with the database.
+operator who cannot fix ownership without database access has been locked out by the thing that
+was supposed to protect them. There are two ways to do it:
+
+```bash
+# From the API, as an administrator — the decommissioned-directory case
+curl -X PATCH "$MLFLOW/api/2.0/mlflow/users/ownership" \
+  -H "Authorization: Bearer $ADMIN_TOKEN" \
+  -d '{"username": "alice@corp.example", "managed_by": "manual"}'
+```
+
+or in bulk with the CLI below, for an operator who has a shell.
 
 ### Changing ownership
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -81,6 +81,7 @@ The plugin uses TTL caches to avoid repeated database lookups on every request. 
 | `OIDC_HTTP_TIMEOUT_SECONDS` | Integer | `10` | Timeout (seconds) applied to OIDC discovery and JWKS HTTP fetches. Set lower for faster failover when the IdP is unreachable; without a timeout a hung IdP can block request threads until the OS-level TCP timeout (~2 minutes), causing cascading auth failures |
 | `OIDC_VERIFY_SSL` | Boolean | `true` | Verify the OIDC provider's TLS certificate on discovery, JWKS, and token requests. Only set to `false` for providers using self-signed certificates in a trusted network |
 | `OIDC_CODE_CHALLENGE` | String | `S256` | PKCE code-challenge method for the authorization-code flow. `S256` (or `true`/`yes`/`on`/`1`), or `none`/`off`/`false`/`no`/`0` to disable. An unrecognised value warns and falls back to `S256`. See [PKCE](#pkce) |
+| `MANAGED_BY_ENFORCEMENT` | String | `report` | What happens when one source writes a row another owns: `off`, `report` (audit only) or `enforce`. See [Row ownership](#row-ownership) |
 | `PERMISSION_CACHE_TTL_SECONDS` | Integer | `30` | Time-to-live (seconds) for the permission resolution cache. Cached permission decisions expire after this duration. Lower values mean faster propagation of permission changes; higher values reduce database load |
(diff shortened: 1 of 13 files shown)


kharkevich:
Filed the known gap as #360: the guard covers `update` only, so `set_user_groups`, `delete` and `create` are unguarded and `user_groups.managed_by` is never read. Group membership is where permissions actually live, so `enforce` currently protects less than it appears to — `report` being the default means nothing is exploitable in a deployment that has not opted in.

Also carried forward into that issue: the #319 review's lesson that lockout scenarios have to be tested *through* the caller. The guard was refusing a directory's own sync because attribution never reached `store.update_user`, and every test passed `written_by=` explicitly, so none of them saw it.

Issue #330, opened 2026-08-09
ui: SAML providers in the login picker


kharkevich:
## Context
Mostly free if #317 was built provider-type-agnostic — which is the reason for doing #317 first.

## Files
- `web-react/src/features/auth/auth-page.tsx` (+ test)

## Scope
- SAML providers appear in the same picker as OIDC providers; the flow is a form POST rather than a redirect.
- Verify the `?next=` return path survives the SAML round trip — it travels in `RelayState`, not in a cookie.

## Acceptance criteria
- [ ] SAML provider appears alongside OIDC providers.
- [ ] Login works from the picker.
- [ ] `next` survives the round trip.

## Depends on
#317, #328.

### Cross-cutting requirements
- **Back-compat**: a deployment that changes no configuration must see no behavior change.
- **Migrations**: forward and backward, tested on SQLite and PostgreSQL.
- **Audit**: identity, provisioning and revocation decisions emit events via `mlflow_oidc_auth/audit.py`.
- **Performance**: must not increase per-request auth query count above the T0 baseline.
- **Conventions**: Black line-length 160, Google/Sphinx docstrings, `snake_case.py`; frontend `kebab-case.tsx` with co-located tests. Conventional Commits, lowercase subject.

---
_Part of the enterprise identity epic #304. Planned in `auth-expansion-research.md` / `auth-expansion-plan.md`._

kharkevich:
Delivered in #373 (merge commit 4676a5c, on `main`).

The picker labels SAML providers with a muted kind hint; OIDC entries render exactly as before. Login stays a plain navigation to the server's `login_url`, now prefixed with the runtime base path when the server could not derive it from a trusted proxy hop.

PR #361, opened 2026-08-12
feat(ui): identity provider picker on the login page (#317)

### LINKED ISSUE #317: ui: identity provider picker on the auth page

kharkevich:
Closes #317.

The login page reads the `/providers` endpoint from #316 and draws one button per identity provider.

**Nothing changes for a single-provider deployment.** With none or one provider it renders exactly the page it always had — the configured label, one button, no picker chrome. A server too old to serve `/providers` reports none and gets that same page, so there is no version coupling.

## What had to be fixed for it to work

The picker was inert against `main`, and three of the four fixes are on the backend:

- **`/providers` was not an unprotected route.** `AuthMiddleware` refused the anonymous fetch, the page saw a non-OK response, fell back to the single-provider login, and sent every visitor to the *default* provider with nothing shown and nothing logged. It is now matched **exactly** rather than by prefix — `/providers` as a prefix would silently unprotect any later route beginning with that string — and asserted end to end through the real middleware in `test_providers_endpoint_e2e.py`, which the existing tests could not see because they call the handler directly.
- **`login_url` no longer carries an origin.** An absolute URL has to get its origin from somewhere, and the only per-request candidate is the `Host` header. It is a `root_path`-relative path now — same-origin by construction, correct under a mount prefix, and needing no configuration to be right. The response is also `Cache-Control: no-store`.
- **The frontend accepts only same-origin paths** — one leading slash, never two, since `//host` is protocol-relative — plus `typeof` checks, because these values become an `href` and a React key and a non-string would be stringified rather than rejected.
- **The fallback button is disabled while the list loads.** It names the *default* provider; a click in the moment before the list arrives would start a login against an IdP the user may not belong to, and with provisioning on, create an account there.

## `?next=`

Carried onto whichever provider is chosen and deliberately **not** validated client-side. `_sanitize_next` owns that decision, it is applied on both `/login` and `/login/{provider_id}`, and a second implementation here would be a second thing to keep correct — starting with the rule that rejects `/\evil.com`.

## Forward-compatible with #330

Nothing in the picker knows what kind of provider it is drawing: the label and the destination both come from the server, and `type` is carried through unused. SAML providers appear without these files changing — there is a test for that.

## Verification

| Command | Result |
|---|---|
| `pytest mlflow_oidc_auth/tests/` (hooks run file by file) | 3574 passed, 14 skipped; hooks 352 passed |
| `pre-commit run --all-files` | passed |
| `yarn test` | 118 files, 857 passed |
| `yarn lint` | no new problems — the 11 errors / 3 warnings on this branch are identical to `origin/main` |

The new `test_providers_endpoint_e2e.py` fails 6/6 without the middleware fix.

### DIFF
--- a/mlflow_oidc_auth/middleware/auth_middleware.py
+++ b/mlflow_oidc_auth/middleware/auth_middleware.py
@@ -129,7 +129,11 @@ class AuthMiddleware(BaseHTTPMiddleware):
             # navigation will redirect through the IdP for re-auth.
             "/static-files",
         )
-        return path.startswith(unprotected_prefixes)
+        # Matched exactly rather than by prefix. A login page has to read this before anyone has
+        # signed in — it is the list of buttons to draw — but "/providers" as a prefix would
+        # silently unprotect any later route whose path merely began with it.
+        unprotected_exact = ("/providers",)
+        return path in unprotected_exact or path.startswith(unprotected_prefixes)
 
     async def _authenticate_basic_auth(self, auth_header: str) -> Tuple[bool, Optional[str], str]:
         """
--- a/mlflow_oidc_auth/routers/auth.py
+++ b/mlflow_oidc_auth/routers/auth.py
@@ -349,28 +349,32 @@ async def providers(request: Request):
                     "id": provider.id,
                     "display_name": provider.display_name or provider.id,
                     "type": provider.type,
-                    "login_url": _absolute_path(request, f"{LOGIN}/{provider.id}"),
+                    "login_url": _login_path(request, f"{LOGIN}/{provider.id}"),
                 }
                 for provider in config.AUTH_PROVIDERS.interactive_providers()
             ]
-        }
+        },
+        # Unauthenticated, and the browser turns each entry into a button it will navigate to.
+        # A shared cache in front of the deployment must not be able to hand one visitor's
+        # response to another.
+        headers={"Cache-Control": "no-store"},
     )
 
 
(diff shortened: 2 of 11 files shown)


kharkevich:
Included in `main` through #373 (merge commit 4676a5c), which carried this branch's two commits unchanged. Closing as merged-via-#373; #317 is closed.

Issue #367, opened 2026-09-04
auth: concurrent silent refresh replays one refresh token and ends the session under rotation with reuse detection


vikranth-t:
## Context

Since 7.16, `_authenticate_session` enforces the IdP-issued `expires_at` and attempts a silent
refresh before rejecting a session
([`middleware/auth_middleware.py:460-472`](https://github.com/mlflow-oidc/mlflow-oidc-auth/blob/main/mlflow_oidc_auth/middleware/auth_middleware.py#L460-L472),
[`routers/auth.py:144`](https://github.com/mlflow-oidc/mlflow-oidc-auth/blob/main/mlflow_oidc_auth/routers/auth.py#L144)).

Against an IdP that rotates refresh tokens and applies reuse detection (OAuth 2.0 Security BCP
§4.14.2 — JumpCloud, Auth0 and Okta with rotation enabled, Keycloak with `revoke-refresh-token`),
that refresh is unsafe to attempt from a browser session.

## What happens

The MLflow UI issues many requests in parallel. They reach the expired session together, so N
requests each enter the `_is_session_expired` branch and each call `refresh_session_with_idp` with
the same stored `refresh_token` ([`routers/auth.py:169`](https://github.com/mlflow-oidc/mlflow-oidc-auth/blob/main/mlflow_oidc_auth/routers/auth.py#L169)).
Nothing serialises them, so each is a separate token exchange presenting an identical token.

The IdP honours the first and treats the other N-1 as replay. Reuse detection then revokes the
whole token family — including the token the winner had just been issued. The next request cannot
refresh, `refresh_session_with_idp` returns `False`, and the middleware clears the session.

The effect is that sessions end roughly once per access-token lifetime (one hour for us). Only
users with more than one request in flight are hit, so it presents as intermittent and is easy to
misread as an IdP problem.

## Why the current code doesn't catch it

- `refresh_session_with_idp` is called straight from the middleware, per request. There is no
  single-flight guard for a given session.
- A failed refresh is treated as "force re-login"
  ([`middleware:465-471`](https://github.com/mlflow-oidc/mlflow-oidc-auth/blob/main/mlflow_oidc_auth/middleware/auth_middleware.py#L465-L471)),
  which is right for a genuinely expired grant but indistinguishable from a revocation the server
  inflicted on itself.
- `_persist_session_auth` writes the rotated token back, so a losing racer can also overwrite the
  winner's valid token with one that has since been revoked.

[...]

PR #368, opened 2026-09-04
feat(auth): allow the IdP token expiry to go unenforced

vikranth-t:
Closes the escape-hatch half of #367. It does **not** fix the underlying race — that
part is left to you, for reasons below.

## What this adds

`OIDC_ENFORCE_IDP_TOKEN_EXPIRY`, default `true`, so existing behaviour is untouched
and no deployment changes on upgrade.

Set it `false` and:

- `_is_session_expired` returns `False` regardless of `expires_at`, so the refresh
  flow in `_authenticate_session` is never entered;
- `_persist_session_auth` stores neither `expires_at` nor `refresh_token`, and
  discards values left by an earlier login;
- the session lives for `SESSION_COOKIE_MAX_AGE_SECONDS` — which since #310 is a
  server-side row and genuinely revocable, not a cookie nobody can withdraw.

The check sits in the middleware as well as at the point the expiry is written, so
sessions established before the setting changed stop being expired immediately,
rather than each needing one more logout for the change to take effect.

## Why

Against an IdP that rotates refresh tokens with reuse detection, refreshing from a
browser session is unsafe: the page issues many requests in parallel, they reach the
expired token together, and each presents the same refresh token. The IdP honours the
first and treats the rest as replay, revoking the chain — including the token the
winner had just been issued. #367 has the full trace.

Against JumpCloud this ended a session roughly every hour, for every user with more
than one request in flight.

## What this deliberately does not do

The real defect is that `refresh_session_with_idp` has no single-flight guard, and
that is yours to design rather than mine to guess. The fix wants a lock on the
`auth_sessions` row (or equivalent), because an in-process lock is not enough for
anyone running multiple replicas — we run several, so a naive asyncio mutex would
not have helped us. That touches your session-store design and possibly the
migration surface, so I would rather not pre-empt it in this PR.

Once the race is fixed, this flag stays useful for deployments that cannot use
refresh tokens at all — but it stops being load-bearing, and we would drop it locally.

## Trade-off

With enforcement off, a change at the IdP (deactivation, group change, MFA
enrollment) no longer takes effect within the token's lifetime. It waits for session
expiry or an explicit revoke here. That is documented on the setting, and the default
keeps the current behaviour, so it is an operator's decision to make knowingly. Pair
it with a shorter `SESSION_COOKIE_MAX_AGE_SECONDS` where that window matters.

## Open question

With enforcement off we still request `offline_access` when `OIDC_USE_REFRESH_TOKEN`
is set, even though nothing will present the refresh token. Happy to make enforcement
off imply not requesting the scope, if you would prefer that — I left the two settings
independent rather than making one silently override the other.

## Testing

Three cases added: the middleware not expiring a session when enforcement is off, and
`_persist_session_auth` both storing neither value and discarding values from an
earlier login when the setting is turned off.

- `pytest mlflow_oidc_auth/tests/middleware/test_auth_middleware.py mlflow_oidc_auth/tests/routers/test_auth.py` — 126 passed
- Full suite — 3913 passed, 14 skipped. The 9 failures in
  `tests/hooks/test_artifact_proxy_coverage.py` reproduce identically on an unpatched
  `main` in my environment (Python 3.10; CI runs 3.12), so they are unrelated to this
  change.
- `black --line-length 160 --check` clean on all touched files.

Running in production on 7.16.0 for the past two weeks.


kharkevich:
Thank you for this, and for the precise diagnosis in #367. #373 (merged as 4676a5c) fixes the underlying defect instead of adding the escape hatch: the refresh token and IdP expiry now live encrypted on the server-side session row rather than in the cookie, refresh is single-flight per session (asyncio lock plus a PostgreSQL row lock, with losers adopting the winner's tokens), and the required Keycloak e2e job reproduces the eight-way parallel refresh with reuse detection on and asserts exactly one refresh at the IdP. With the race gone the flag would no longer be load-bearing, so it is not carried; the pre-#367 behaviour it restored (sessions living for the cookie lifetime with no IdP check) is deliberately not offered as a switch. Closing without merge.

PR #370, opened 2026-09-10
fix(auth): bridge AuthContext to FastAPI routes via ContextVar

### LINKED ISSUE #369: [BUG] FastAPI-native routes return 403 for users with workspace-only permissions
> **Security vulnerability?** Do not describe it here. Follow [SECURITY.md](../../SECURITY.md):
> open a stub titled `[BUG] Security Vulnerability` with **no details**, then email the evidence
> to `opensource-security@kharkevich.com`.

## What happened
`POST /v1/traces` returns 403 for non-admin users whose only permission path is workspace-level MANAGE (direct or regex).

## What you expected
Workspace MANAGE grants experiment UPDATE, which is what the OTel validator checks - so the request should be authorized.

## Reproduction

Steps, or a minimal configuration that triggers it.

```
1. Set MLFLOW_ENABLE_WORKSPACES=true, DEFAULT_MLFLOW_PERMISSION=NO_PERMISSIONS
2. Create workspace "team-ws" with an experiment in it
3. Grant a group workspace MANAGE on "team-ws"
4. Log in as a non-admin user in that group with no experiment-level permissions
5. POST /v1/traces with X-Mlflow-Experiment-Id and X-MLFLOW_WORKSPACE: team-ws -> 403 Forbidden

curl -i -X POST "https://<host>/v1/traces" \
  -u "<user>:<pass>" \
  -H "X-Mlflow-Experiment-Id: <experiment-id>" \
  -H "X-MLFLOW-WORKSPACE: team-ws"
```

The same user succeeds on Flask-routed APIs (e.g. `GET /api/2.0/mlflow/experiments/get`) which get the `AuthContext` via `AuthAwareWSGIMiddleware`.
```
curl -i -u "<user>:<pass>" \
  -H "X-Mlflow-Experiment-Id: <experiment-id>" \
  -H "X-MLFLOW-WORKSPACE: teamws"
  "https://<host>/api/2.0/mlflow/experiments/get?experiment_id=<experiment_id>"
```

## Environment

- `mlflow-oidc-auth` version: 7.18.1
- MLflow version: 3.14+
- Python version: 3.12
- Database (SQLite / PostgreSQL / MySQL): PostgreSQL
- Identity provider (Keycloak / Entra ID / Okta / other): EntraID
- Deployment (single replica / multi-replica, behind a proxy?): single replica

## Logs

Relevant log output. **Redact tokens, secrets, and full JWTs** — a JWT payload is base64, not
encryption.

```
HTTP/1.1 403 Forbidden
Permission Denied

INFO: "POST /v1/traces HTTP/1.1" 403 Forbidden
```

## Additional context

Admins are unaffected (early bypass). Users with direct experiment permissions are unaffected (resolved before fallback). `_get_cache_workspace()` is affected the same way, so cache keys on these routes always use the `::no-workspace` marker.

henryinqz:
## Summary

`get_auth_context()` only reads Flask's `request.environ`. FastAPI-native routes never enter Flask, so it always raises -> `get_request_workspace()` returns `None` -> `apply_workspace_fallback() is skipped` -> `DEFAULT_MLFLOW_PERMISSION` applies.
Users whose only grant is workspace-level MANAGE get 403 on `/v1/traces

The `AuthContext` is already in the ASGI scope at `request.scope["mlflow_oidc_auth"] (set by `AuthMiddleware`); it just never reaches the bridge.

Closes #369

## Changes

`bridge/user.py` - add `_auth_context_var` with `set_auth_context()`/`clear_auth_context()`. `get_auth_context()` tries Flask environ first, then falls back to the ContextVar.

`middleware/fastapi_permission_middleware.py` - set the `AuthContext` from the ASGI scope before running the validator; clear it in `finally`.
Mirrors the existing `WorkspaceContextMiddleware` ContextVar pattern.

## Security
- Deny-by-default preserved - every failure path returns `None`/raises/`False`
- `ContextVar` is per-`asyncio.Task`, so no cross-request leakage; `finally` adds defence in depth
- Flask environ keeps precedence - no bypass
- No new `is_admin` path - admin bypass reads `request.state` and returns before `set_auth_context`
- No new DB queries on the auth path; `AuthContext` is a frozen dataclass

## Test plan
- [x] `pre-commit run --all-files` - 8/8 hooks pass, no files rewritten
- [x] `pytest mlflow_oidc_auth/tests/bridge/test_user.py` - 44 passed, incl. 8 new (ContextVar fallback, cleanup, Flask-environ precedence)
- [x] `pytest mlflow_oidc_auth/tests/middleware/test_fastapi_permission_middleware.py` - 56 passed, incl. 4 new (workspace propagation into the validator, cleanup on success and on exception, no-workspace case
- [x] `pytest mlflow_oidc_auth/tests/utils/test_permissions.py -k workspace` - 9 passed, existing workspace-fallback behaviour unchanged
- [x] `pytest mlflow_oidc_auth/tests/ --ignore=mlflow_oidc_auth/tests/hooks` - 3588 passed, 14 skipped, 1 failed (see below) 
- [x] `hooks/` run file-by-file per AGENTS.md's documented hang - 9 of 10 files clean

**Two pre-existing failures, unrelated to this change.** Both reproduce identically on the parent commit `76a6dd0` with none of this PR's code::
- `test_app.py::test_mcp_server_registry_router_included` - asserts `/api/3.0/mlflow/mcp-servers` is registered; absent in MLflow 3.14.0, the declared floor
- `hooks/test_artifact_proxy_coverage.py` - 9 failed, 57 passed
 
- [ ] Manual: workspace-MANAGE-only user POSTs `/v1/traces` -> 200

### DIFF
--- a/mlflow_oidc_auth/bridge/user.py
+++ b/mlflow_oidc_auth/bridge/user.py
@@ -3,22 +3,44 @@ Flask Hooks Bridge - Compatibility Layer for Flask Hooks with FastAPI Auth
 
 Provides functions to retrieve authentication context from Flask's WSGI environ,
 where it was injected by AuthAwareWSGIMiddleware from FastAPI's ASGI scope.
+
+For FastAPI-native routes (OTel traces, gateway, etc.) that never enter Flask,
+a ContextVar fallback allows the same bridge functions to work when the
+FastAPI permission middleware sets the AuthContext before running validators.
 """
+
+from contextvars import ContextVar
 
 from mlflow_oidc_auth.entities.auth_context import AUTH_CONTEXT_KEY, AuthContext
 from mlflow_oidc_auth.logger import get_logger
 
 logger = get_logger()
 
+_auth_context_var: ContextVar[AuthContext | None] = ContextVar("_auth_context_var", default=None)
+
+
+def set_auth_context(ctx: AuthContext) -> None:
+    """Set AuthContext in the ContextVar for non-Flask contexts (FastAPI-native routes)."""
+    _auth_context_var.set(ctx)
+
+
+def clear_auth_context() -> None:
+    """Clear the ContextVar after request processing."""
+    _auth_context_var.set(None)
+
 
 def get_auth_context() -> AuthContext:
-    """Get the full AuthContext from Flask request environ.
+    """Get the full AuthContext from Flask request environ or ContextVar fallback.
+
+    Checks Flask's WSGI environ first (for routes that pass through
+    AuthAwareWSGIMiddleware), then falls back to the ContextVar (for
+    FastAPI-native routes where the permission middleware sets it).
 
     Returns:
         AuthContext object containing username, is_admin, and workspace.
 
     Raises:
-        Exception: If AuthContext is not available in the environ.
+        Exception: If AuthContext is not available from either source.
     """
     try:
         from flask import request
@@ -31,11 +53,16 @@ def get_auth_context() -> AuthContext:
     except Exception as e:
         logger.debug(f"Could not access AuthContext from Flask request: {e}")
 
+    ctx = _auth_context_var.get()
+    if isinstance(ctx, AuthContext):
+        logger.debug(f"Retrieved AuthContext from ContextVar: {ctx.username}")
+        return ctx
+
     raise Exception("Could not retrieve AuthContext")
 
 
 def get_fastapi_username() -> str:
-    """Get username from FastAPI authentication context via Flask request environ.
+    """Get username from the current authentication context.
 
     Returns:
         Username if authenticated.
@@ -53,7 +80,7 @@ def get_fastapi_username() -> str:
 
 
 def get_fastapi_admin_status() -> bool:
-    """Get admin status from FastAPI authentication context via Flask request environ.
+    """Get admin status from the current authentication context.
 
     Returns:
         True if user is admin, False otherwise.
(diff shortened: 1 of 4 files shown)


henryinqz:
I have updated the PR to address previous CI failures

kharkevich:
Reviewed and validated locally (pre-commit, `tests/bridge`, `tests/middleware`, `tests/utils/test_permissions.py`, and the auth-path query baseline, plus a read-only security pass). The design is right: the bridge carries the raw normalised header (None when absent), which is what `_apply_workspace_fallback` documents and what `_get_cache_workspace()` already reads, so cache key and fallback now agree on FastAPI routes. No cross-request leakage path: the var is reset before `call_next`, and admin/401 paths return before it is set. Zero added statements on the auth path.

Pushed one follow-up commit (93db6fd):

- `clear_auth_context` now resets the ContextVar with the token from `set_auth_context` rather than blanking it, so it restores the prior value instead of clobbering any outer setter.
- `test_contextvar_cleared_after_request` / `_on_validator_exception` read the var from the TestClient thread, which has its own context and always sees None, so they passed with the `finally` removed. They now observe it from inside the app and fail under that mutation.
- Added `TestWorkspaceFallbackEndToEnd`: the real `effective_experiment_permission` → `_apply_workspace_fallback` path on `/v1/traces` with only store lookups stubbed. Workspace MANAGE → 200, no grant on the request workspace → 403, workspace READ → 403, and a regression guard with the bridge disabled → 403. That is the #369 scenario end to end, including the negative paths.

One pre-existing note, no change made: `X-MLFLOW-WORKSPACE` is client-chosen and now drives authorization on `/v1/traces` and `/gateway/*` the same way it already does on Flask routes. Containment relies on MLflow scoping the store to the same header via `WorkspaceContextMiddleware`.

kharkevich (approved):

PR #371, opened 2026-09-21
perf(hooks): skip the protobuf round-trip when a logged-model page needs no filtering

lazarevtill:
## What this changes

`_filter_search_logged_models` parses the response body into protobuf, drops the models the
caller may not read, then re-serialises the whole message — and it does that last step
unconditionally, even when it removed nothing.

That is expensive here in a way it is not for the other filters, because a page of logged models
carries every metric value of every model on it, so the body is routinely several megabytes.

This decides on the raw JSON whether anything has to change. When every model is readable, and the
tracking server reported no further page to top the current one up from, the body it produced is
already the correct answer and is passed through untouched.

## Measurements

One page of a real experiment: 28 models holding 20,720 metric values, 7.15 MB of JSON. The hook
alone, best of five:

| | time |
|---|---|
| today (`parse_dict` + re-encode) | 484.1 ms |
| with the pass-through | **15.8 ms** |

For context the tracking server spends ~616 ms producing that response, so the hook is currently
about 43% of the request and in the common case it changes nothing.

Breakdown of the 484 ms: 17 ms `resp.json`, 214 ms `parse_dict`, 242 ms re-encoding.

## Why this is safe

The shortcut is deliberately one-sided. `_search_logged_models_needs_no_filtering` returns True
only when every model has been positively confirmed visible; anything it cannot verify returns
False and falls through to the existing filter unchanged:

- a body that is not an object, or a `models` value that is not a list
- a model without a string `info.experiment_id`
- a workspace it cannot resolve, when `MLFLOW_ENABLE_WORKSPACES` is on (WSSEC-03)
- a non-empty `next_page_token`, since the filter may then need to top the page up

It cannot pass through a model the old code would have removed.

## One visible difference

When the shortcut fires, `next_page_token` is absent from the body rather than the empty string
the old path always wrote. That matches what the tracking server itself emits when there is no
next page.

## Tests

Nine new tests in `test_after_request.py`: the pass-through (asserting the body is not rewritten),
an unreadable model, a present next page, workspace denial, and five malformed-body shapes.

Three existing tests set `resp.json` to a body that disagrees with the protobuf message they mock
`parse_dict` into producing, which cannot happen when `parse_dict` actually runs. Their bodies now
say what their messages claim, so they keep exercising the paths they were written for. Their
assertions are unchanged.

Full suite before this change: 18 failed, 3911 passed, 14 skipped. After: 18 failed, 3920 passed,
14 skipped — the same failures (all pre-existing: artifact-proxy `file:` URIs, db-utils
migrations, hack/index), with the passes up by exactly the nine tests added. `black --check` clean.


kharkevich:
Reviewed and validated locally (black, pre-commit, `test_after_request.py` and `test_workspace_search_filtering.py` run file by file, plus a read-only security pass on the filter). The shortcut is fail-closed: I could not construct a body where it returns True but the protobuf path would have removed a model, and the benchmark reproduces the speed-up (~230 ms vs ~8 ms on a synthetic 28-model page).

Pushed one follow-up commit (57ee9be) for three low-severity items:

- `Response.get_json` is not cached by werkzeug, so when filtering *was* needed the body got decoded twice. The payload is now decoded once and handed to both the shortcut and `parse_dict`.
- With workspaces on, the shortcut's `get_experiment` lookups were thrown away when it bailed, and the full filter repeated them. It now fills the same `exp_ws_map`.
- `test_filter_search_logged_models_no_pagination_needed` and `test_keeps_logged_models_from_accessible_workspaces` used `{"models": []}`, which now takes the pass-through and left their assertions vacuous. Given a next-page token like their siblings.

Behaviour of the shortcut itself is unchanged.

kharkevich (approved):

